PR #15, opened 2025-11-24
fix: the rest of the dino

hanthor:


system_files/usr/bin/ugum
castrojo: We don't need this.

castrojo (changes requested):
We should probably cherry pick?

PR #32, opened 2025-12-09
feat: add full-desktop Brewfile with GNOME Circle apps

castrojo:
Ok so here's the deal, I want to ship a full blown desktop. Like, if you had a friend that you wanted to experience the very best of local apps, add it here. These will be GNOME Apps, this is more of "Let's ship a full GNOME App experience". 

I'm starting with the GNOME Circle Apps, and I'd like to add Collabora Office, and any other S-tier GNOME App. 

Eventually we'll find a way to add this to first run and/Bazaar's curated section so that we advertise this as a feature.

## References
- Homebrew flatpak support: https://github.com/Homebrew/brew/pull/21097
- GNOME Circle: https://circle.gnome.org/


renner0e:
Can this please be moved to be bluefin specific? Thanks

hanthor (approved):

PR #47, opened 2025-12-14
fix: remove resolve ujust recipe

castrojo:
This gives off the impression that we do resolve support, so removing it.


renner0e (approved):
if it's broken might as well remove it

PR #66, opened 2025-12-19
feat(just): add clean-system back

renner0e:


system_files/shared/usr/share/ublue-os/just/default.just
```diff
@@ -25,6 +25,17 @@ bios-info:
     echo "Release Date: $(dmidecode -s bios-release-date)"
     EOF
 
+clean-system:
+    #!/usr/bin/bash
+    podman image prune -af
+    podman volume prune -f
```
castrojo: Can we do an if statement here if docker exists to do the system prune there as well? 

We should add a gum confirmation as well before since we're -f'ing the prunes.

renner0e: https://github.com/projectbluefin/common/pull/66/changes/371c8d81829d1435df6aab7f87c52958ad14d1ec

castrojo (approved):

PR #74, opened 2025-12-24
fix: re-include zsa udev rules

tulilirockz:
Someone on discord specifically asked for these. We had removed a few
udev rules before since we wanted to trim down our dependencies and
these seemed to be quite niche udev rules that maybe we could remove.
Turns out someone noticed it and we are re-adding them now :)


inffy (approved):
LGTM

PR #92, opened 2025-12-29
chore(ujust, bluefin): do not run rollback helper on centos

tulilirockz:
We should re-enable it once we properly handle it on LTS


ledif (approved):
LGTM. Gemini's suggestion of adding `-q` is okay but it's fine as is without change.

renner0e (approved):

PR #95, opened 2025-12-29
feat: add Cryptomator to utilities list

sebjag:
Not sure if there is a better place to make suggestions like this, but here's a suggested app to consider adding (if you would like :smiley:).

This app makes it easy to create password protected and encrypted folders:  https://docs.cryptomator.org/desktop/getting-started/


castrojo (approved):
official app, SGTM, thanks!

PR #98, opened 2025-12-30
feat: add Jan flatpak to ai-tools Brewfile

coxde:
[Jan](https://github.com/janhq/jan) is a similar tool like LM Studio, but with Apache 2.0 license, and it's already on Flathub. Might be worth to add to the list!


coxde:
> Would you mind adding it to the curated section in bazaar too? We only have two apps there, this is a nice find, thanks!
> 
> image

Oh yeah absolutely, it came to me the second I submitted this pr lamo, will do in another pr!

castrojo (approved):

PR #154, opened 2026-01-09
feat: Add translation support for motd

theMimolet:
Hi ! 

I was wondering to myself if it was possible to add translations to the motd. 
So I tested a bit, and here you go ! :+1: 

It's still a draft PR because I still need to fully translate it in french (ofc) and I'm not sure about my changes in env.sh.

Any feedback is welcome ! :smile: 
_(Of course, if it works out here, it will also be added to Bazzite and Aurora)_


theMimolet:
Here's what the updated version contains : 
- a new /lang folder for translations, it uses .json files (like the website)
- the /tips folder now also contains the tips in different languages
- ublue-motd has seen big additions to support the string based translations

Basically it will get the user's language code, check if a file for it already exist, retrieve the strings, retrieve one random tip (from the user's language code) and then finally renders the motd.

system_files/shared/usr/bin/ublue-motd
```diff
@@ -1,8 +1,30 @@
 #!/usr/bin/env sh
 
-MOTD_TEMPLATE_FILE="${MOTD_TEMPLATE_FILE:-/usr/share/ublue-os/motd/template.md}"
-. "${MOTD_ENV_SCRIPT:-/usr/share/ublue-os/motd/env.sh}"
+. "${MOTD_ENV_SCRIPT:-./env.sh}"
```
ledif: I don't understand why this line changed from 

```
. "${MOTD_ENV_SCRIPT:-/usr/share/ublue-os/motd/env.sh}"
```

ledif (commented):
At a quick glance, this looks good. I haven't tested it, but it seems like it should work. I'd be more comfortable approving if someone could test it.

system_files/shared/usr/bin/ublue-motd
theMimolet: OH GOSH, GOOD CATCH
I forgor to put it back !

PR #161, opened 2026-01-10
fix(mise): Auto install and activate mise unless told otherwise

rwaltr:
Simple mise integration

Enables auto mise activation if installed. (includes it in cli brewfile)

Can disable by setting `MISE_[SHELL]_AUTO_ACTIVATE=0`

If mise is installed, enables packages to be used for the user by running `mise -g use [package]`, then running `mise install`


tulilirockz (changes requested):
We should probably make this a part of bling instead IMHO

tulilirockz (approved):

castrojo (approved):

PR #169, opened 2026-01-11
Lets add mise to development

rwaltr:
This intentionally does not replace the current github actions and justfiles.

They aint broke yet.

This just eases in the tools used as mise config, and the recommended gitignores.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+mise.local.toml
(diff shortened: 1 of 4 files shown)


castrojo (approved):

tulilirockz (changes requested):
As we are discussing on discord, we'd probably need a similar PR to this to all the Bluefin/(ublue?)-related repos to this since we use this kind of tooling everywhere. But even then we have `bluefin-cli`, which includes all of these and `podman`/`cosign`/`just` are already on bluefin itself. 

I'm thinking this could be a really interesting idea though, what if we made a brewfile for _contributing_ to bluefin? "Hey, do you want to contribute? Install our recommended packages and get to work!" I think bbrew/brew bundle can access remote URLs probably.

rwaltr:
Closing as mise in the image and mise as development are two different things.

need to establish documentation for mise first.

PR #183, opened 2026-01-23
feat: Add org.gnome.SoundRecorder to system-flatpaks

LorbusChris:
Add this handy GNOME app for voice memo recording.


castrojo (approved):
Nice catch!

tulilirockz (approved):

PR #202, opened 2026-02-03
fix: move ublue signing keys to /usr/lib/pki/containers

renner0e:
This is more FHS compliant and more prone to people just deleting random stuff in /etc.

If something is wrong with these files then updates will break!

This needs to be really well tested by as many people as possible 

Our tests should catch this in the build

https://github.com/ublue-os/aurora/blob/614675502eb9d1c22622bb5c506d73335a862002/build_files/base/20-tests.sh#L13-L16


inffy (approved):

castrojo (approved):

m2Giles:
For context on why this was placed in /etc

When we had the breakage, everyone was pointed to a /usr/etc path. Since we couldn't have people replace the public key in /usr/etc we had to modify policy.json file.

Since the breakage we can now support multiple signing keys so this is less of a risk. But changing to a writeable location was purposeful.

PR #263, opened 2026-04-06
feat: enable GNOME Calculator search provider by default

dylanmtaylor:
Adds org.gnome.Calculator.desktop to the enabled search providers list so users can do math directly from GNOME search without manually enabling it in settings.

Tested locally by applying the gsettings change and confirming calculator results appear inline when typing math expressions in the GNOME overview search.

image

Closes https://github.com/projectbluefin/common/issues/256


dylanmtaylor:
You can also test with gsettings set org.gnome.desktop.search-providers enabled "['io.github.kolunmi.Bazaar.desktop', 'org.gnome.Calculator.desktop']"

Then validate by typing a math expression into the search bar.

castrojo (approved):

castrojo (approved):

ahmedadan (approved):

Issue #222, opened 2026-02-27
Investigate Custom Command Menu instead of Logomenu


castrojo:
Our translation story for logomenu has always not worked, but then I saw this on the gnome newsletter and tried it. 

https://github.com/StorageB/custom-command-menu

I like it:  

<img width="339" height="528" alt="Image" src="https://github.com/user-attachments/assets/930f9ac5-6215-48d1-b15b-ef4431f842c4" />

We need to figure out if we can translate this, and also if we can figure out how to put the hostname there I think that's fine. We'd need to figure out the icon in the panel and see if we can put the `u` there. 

Also let's not do symbolic icons next to the menu items that'll get cluttered. Thoughts?

castrojo:
Sent to you in discord, can't attach here. Was also thinking how we can pull off the hostname. I was wondering if we could just set the settings key to be the current hostname or something? It'd be awesome if we could do the hostname.

PR #264, opened 2026-04-06
feat: replace LogoMenu with Custom Command Menu extension

dylanmtaylor:
Replaces logomenu with [Custom Command Menu](https://github.com/StorageB/custom-command-menu). Supports translations, actively maintained, fully customizable entries.

The main bluefin image Containerfile will also need to swap the extension package.

image

Ref: https://github.com/projectbluefin/common/issues/222


castrojo:
The menu doesn't look like anything specced in the issue? We just need the one in the issue implemented. And we ship distroshelf not boxbuddy.

castrojo:
We can punt on "Ask Bluefin" until latest though so we can leave that one out, thanks!

dylanmtaylor:
> The menu doesn't look like anything specced in the issue? We just need the one in the issue implemented. And we ship distroshelf not boxbuddy.

@castrojo Odd, my bluefin has boxbuddy instead of distroshelf. I pushed a version with your menu (incl. ask bluefin), and came up with a way to get the hostname in the menu using a oneshot service that starts on graphical session start. There's no way to evaluate within custom menu.

dylanmtaylor:
image

castrojo:
+1 from me, thanks for making the adjustments, since this is a service I'll let the others comment.

dylanmtaylor:
Let's see if we can get https://github.com/StorageB/custom-command-menu/pull/10 upstreamed and then use that for the hostname setting

castrojo:
Can we land the config now so we can ship custom menu in dakota? Since we're making good progress there we might as well ship it.

dylanmtaylor:
> Can we land the config now so we can ship custom menu in dakota? Since we're making good progress there we might as well ship it.

I'm happy to land this now and pull out the systemd service and replace it with the new version in the future.

castrojo (approved):

ahmedadan (approved):

PR #272, opened 2026-04-13
feat: use dynamic hostname in menu label

dylanmtaylor:
Now that StorageB/custom-command-menu#10 added support for `$(command)` substitution in labels, replace the hardcoded "Bluefin" section header with `$(hostname)` so the menu displays the machine's actual hostname.

Follow-up to #264.


dylanmtaylor:
Note that while this PR is in its final state, we need to wait for the new version to be tagged and pushed to GNOME extensions before merging this in.

castrojo (approved):

PR #275, opened 2026-04-14
feat: add Files to custom command menu

dylanmtaylor:
Add Files (Nautilus) entry above Terminal in the custom command menu, so users can access the file manager even if the dock is disabled.


castrojo:
Not going with this one, we can reinvestigate menu options later.

PR #291, opened 2026-04-23
feat(ublue-bling): clean up fish script

AlexNPavel:
Add a variety of fixes to the bling.fish script to bring it up to par with bling.sh.

- only blingify in an interactive terminal
- add no-source-twice fix
- Simplify command existence checks
- Add command existence check for bat
- Add direnv and mise
- Use `command init fish | source` pattern (every tool has that pattern as the correct way to set up the tool in their respective documentations).
- Use `abbr -a` instead of `alias` by default. Abbr is generally preferred over alias in fish. This does result in a functional change though: `abbr` expands in the terminal. This is one of the reasons they are preferred (you can see the exact command instead of the real command being effecively hidden), but some people may prefer the simpler aliases, so I added a flag similar to what `mise` has to disable `abbr` and go back to `alias` if a user prefers that.

### DIFF
--- a/system_files/shared/usr/share/ublue-os/bling/bling.fish
+++ b/system_files/shared/usr/share/ublue-os/bling/bling.fish
@@ -1,35 +1,45 @@
 #!/usr/bin/env fish
 
-# FIXME: add no-source-twice fix
-
-# ls aliases
-if [ "$(command -v eza)" ]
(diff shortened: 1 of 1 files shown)


castrojo:
I don't use fish but this seems sane to me, thank you!

castrojo (approved):

inffy (approved):

PR #300, opened 2026-05-04
fix: ignore commented LockLayering setting

joshyorko:
## Summary
- only treat active `LockLayering=false` as enabled
- keep commented examples from selecting `rpm-ostree upgrade`

## Validation
- checked commented `# LockLayering=false` does not match
- checked active `LockLayering=false` still matches

### DIFF
--- a/system_files/shared/usr/share/ublue-os/just/update.just
+++ b/system_files/shared/usr/share/ublue-os/just/update.just
@@ -14,7 +14,7 @@ update:
         echo "automatic updates are currently running, use \`journalctl -fexu $SERVICE\` to see logs"
         exit 1
     fi
-    if grep -q -E -e "LockLayering=false" /etc/rpm-ostreed.conf ; then
+    if grep -q -E -e "^[[:space:]]*LockLayering[[:space:]]*=[[:space:]]*false([[:space:]]*(#.*)?)?$" /etc/rpm-ostreed.conf ; then
         rpm-ostree upgrade
     else
         sudo bootc upgrade


castrojo (approved):

system_files/shared/usr/share/ublue-os/just/update.just
```diff
@@ -14,7 +14,7 @@ update:
         echo "automatic updates are currently running, use \`journalctl -fexu $SERVICE\` to see logs"
         exit 1
     fi
-    if grep -q -E -e "LockLayering=false" /etc/rpm-ostreed.conf ; then
+    if grep -q -E -e "^[[:space:]]*LockLayering[[:space:]]*=[[:space:]]*false([[:space:]]*(#.*)?)?$" /etc/rpm-ostreed.conf ; then
```
ehsanullahjan, line 17: this regex is robust, but i wonder if we could simplify it? two simplified variations i could think of...

if we don't wanna cover _all_ possible ways this line might be formatted, simply this would do:
`if grep -q "^LockLayering=false" /etc/rpm-ostreed.conf; then`

if we intend to be comprehensive, this simpler variation is more than sufficient:
`if grep -Eq "^\s*LockLayering\s*=\s*false" /etc/rpm-ostreed.conf; then`.

castrojo (approved):

hanthor (approved):
Correct regex fix -- only matching active LockLayering=false and not commented-out examples is exactly right. CI passes. Approved.

hanthor (approved):
Correct fix -- only matching active LockLayering=false and not commented-out examples is right. CI passes. Approved.

PR #324, opened 2026-05-14
fix: removing search-light from the enabled extensions until there's a fix

pfanzola:
Consider to disable the search-light extension until they'll fix this:  https://github.com/ublue-os/bluefin/issues/4622
https://github.com/icedman/search-light/issues/161

Normally, I would wait for a fix from upstream (without this PR), but this bug can log the user out or even cause the computer to reboot, resulting in the loss of hours of unsaved work.

## Summary by CodeRabbit

* **Chores**
  * Disabled the "Search Light" GNOME extension from the system default configuration so it will no longer be enabled by default. This changes the out-of-the-box desktop experience by preventing the extension from loading automatically; users may still enable or install it manually if they prefer.

[Review Change Stack](https://app.coderabbit.ai/change-stack/projectbluefin/common/pull/324?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


castrojo:
I think let's at see least if we can find someone to fix this upstream?

pfanzola:
> I think let's at see least if we can find someone to fix this upstream? 

I would agree with u but it seems that the search light extension isn't well actively maintained... The last merged PR was 2 mo ago (gnome 50 support) and before that only the 49 support 8 mo ago with some PRs still open ready to be reviewed and mergerd for several months...

Don't get me wrong I would love to see a fix for this instead of cut it off... in fact my PR was only for disabling it, not for the removal of the extension.
Anyway let's wait a bit and see what will happen in upstream.

hanthor (approved):
🤖 **AI-Assisted Pull Request Review**

Fully support this change. Disabling the `search-light` extension is a necessary hotfix to prevent the severe upstream bug that triggers sudden user logouts and system reboots, preventing loss of unsaved user work.

**Concerning CodeRabbit's Nitpick:**
CodeRabbit recommended adding comments explaining why the settings blocks are retained inside `@system_files/bluefin/usr/share/glib-2.0/schemas/zz0-bluefin-modifications.gschema.override`. 
However, **this is actually bad practice and should be avoided**. Comment lines (e.g. starting with `#`) inside `.gschema.override` files are parsed by `glib-compile-schemas` and have historically triggered syntax compilation warnings or errors depending on the specific GLib version on the host. It is much safer to merge this clean diff as-is without adding override comments.

Ready to merge!

hanthor (approved):
Correct fix -- disabling search-light until the upstream crash/logout bug is fixed is the right call for user safety. CI passes. Approved.

hanthor (approved):
Approved: Clean docs changes.

castrojo (approved):

oameye:
This PR should fix it
https://github.com/icedman/search-light/pull/164

PR #333, opened 2026-05-18
feat(bazaar): add LM Studio to AI section

KiKaraage:
https://flathub.org/apps/ai.lmstudio.lm-studio

It has passed the update test, the Flatpak version have good parity with other platforms now

## Summary by CodeRabbit

* **New Features**
  * Added a new AI and Machine Learning application to the curated recommendations list.

[Review Change Stack](https://app.coderabbit.ai/change-stack/projectbluefin/common/pull/333?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/system_files/bluefin/etc/bazaar/curated.yaml
+++ b/system_files/bluefin/etc/bazaar/curated.yaml
@@ -517,3 +517,4 @@ rows:
             - ai.jan.Jan
             - ink.whis.Whis
             - io.speedofsound.SpeedOfSound
+            - ai.lmstudio.lm-studio


ehsanullahjan (approved):

castrojo:
We have this in brew already?

KiKaraage:
Aah yeah apparently the submitter is a random person too. I guess we can either close this or leave this open in case LM Studio verified it someday?

castrojo:
Yeah and the brew one has been pretty solid, feels like this would be a sidegrade.

ehsanullahjan:
is the brew formula based on lm studio appimage by any chance? lm studio flatpak was dropped from default dev flatpaks because it wasn’t being updated. but, looks like it’s up to date now (in fact, the flatpak was slightly newer than appimage when i checked yesterday).

totally fine if we don’t recommend the flatpak, but i thought the context might be helpful.

KiKaraage:
> is the brew formula based on lm studio appimage by any chance? lm studio flatpak was dropped from default dev flatpaks because it wasn’t being updated. but, looks like it’s up to date now (in fact, the flatpak was slightly newer than appimage when i checked yesterday).
> 
> totally fine if we don’t recommend the flatpak, but i thought the context might be helpful.

Yes as you can see [here](https://github.com/ublue-os/homebrew-tap/blob/main/Casks/lm-studio-linux.rb) - though both are in the latest version too

castrojo:
As discussed in chat we'll just stick with homebrew for now but would totally switch to an official flatpak.

PR #352, opened 2026-05-29
fix(ujust): reload udev and unload drivers after OpenTabletDriver install

### LINKED ISSUE #340: OpenTabletDriver ujust does not properly blacklist kernel drivers
```
Common 🧱  ·  issue pipeline
─────────────────────────────────────────────────
  ✓  filed      report received
  ✓  approved   signed off by a maintainer
  ✓  queued     —
  ✓  claimed    —
  ▶  done       fix shipped
─────────────────────────────────────────────────
  report:       missing     ·  verified: 0/3
  area:         just        ·  priority: —
  next action:  ujust verify 340 — three verifies closes the case
```
🦖 rawr

---https://opentabletdriver.net/Wiki/Documentation/RequiredPermissions#package-manager
﻿
I need to do some more testing but I believe the issues ive been running into with art tablets is actually because the ujust script for opentabletdriver seems to only update the udev rules, but its not properly blacklisting wacom and hid_uclogic, 
﻿
I think not blacklisting leads to a confusing and frustrating experience for users since both settings are constantly conflicting. I didnt realize this wasn't "normal" because there was usually a combination I could find that seems to work normally. This is behavior I have dealt with since I started using Bluefin, Aurora, and Bazzite. 
﻿
Running this script I got off the OTD website resolves the issue completely after a reboot
﻿
```
#!/bin/bash
﻿
echo "Finding old udev rules..."
for c in /etc/udev/rules.d/9{0,9}-opentabletdriver.rules; do
  if [ -f "${c}" ]; then
    echo "Deleting ${c}"
    sudo rm "${c}"
  fi
done
﻿
echo "Finding old kernel module blacklist rules..."
if [ -f /etc/modprobe.d/blacklist.conf ]; then
  echo "Deleting /etc/modprobe.d/blacklist.conf"
  sudo rm /etc/modprobe.d/blacklist.conf
fi
﻿
sudo modprobe uinput
sudo rmmod wacom hid_uclogic > /dev/null 2>&1
﻿
sudo udevadm control --reload-rules && sudo udevadm trigger
```
﻿
﻿
Im willing to PR something for this myself, but I don't know if itd be better to include an extra step in the current ujust "do you want to disable the built in tablet drivers, you will need to edit tablet options in OTD instead of system settings". Or to just rewrite the current one completely.
﻿
I also believe its possible that the udev rules could slowly get messed up when the OTD flatpak updates, so an option to just "fix OTD" could be nice rather then having to uninstall and reinstall with the ujust

hanthor:
## Summary

The `install-opentabletdriver` recipe wrote udev rules and a kernel module blacklist but never applied them, requiring a reboot before OTD could take control of the tablet. Additionally the uninstall path tried to remove `blacklist-opentabletdriver.rules` (wrong extension) instead of `blacklist-opentabletdriver.conf`, leaving the blacklist file in place after uninstall.

**Install path:**
- Remove stale `90/99-opentabletdriver.rules` from prior installations before copying the new `71-opentabletdriver.rules` (avoids conflicting rule sets)
- Run `udevadm control --reload-rules && udevadm trigger` to activate new rules immediately
- `modprobe uinput` to load the uinput module OTD requires
- `rmmod wacom hid_uclogic` (best-effort, `|| true`) to unload conflicting drivers — users no longer need to reboot

**Uninstall path:**
- Fix filename: `blacklist-opentabletdriver.rules` → `blacklist-opentabletdriver.conf`
- Run `udevadm control --reload-rules && udevadm trigger` after removing rules

Closes #340

## Test plan
- [ ] Install OTD on a machine with a Wacom or UC-Logic tablet — tablet should work without rebooting
- [ ] Uninstall OTD — verify `blacklist-opentabletdriver.conf` is removed from `/etc/modprobe.d/`
- [ ] Verify `71-opentabletdriver.rules` is removed from `/etc/udev/rules.d/`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/system_files/shared/usr/share/ublue-os/just/apps.just
+++ b/system_files/shared/usr/share/ublue-os/just/apps.just
@@ -30,9 +30,15 @@ install-opentabletdriver:
         | tar --strip-components=1 -xvzf - -C "${OTD_TMPDIR}"
 
       # https://opentabletdriver.net/Wiki/Documentation/RequiredPermissions
+      # Clean up any old rules files from previous installs that may conflict
+      sudo rm -f /etc/udev/rules.d/90-opentabletdriver.rules /etc/udev/rules.d/99-opentabletdriver.rules
       sudo cp "${OTD_TMPDIR}/etc/udev/rules.d/70-opentabletdriver.rules" /etc/udev/rules.d/71-opentabletdriver.rules
       echo -ne "blacklist hid_uclogic\nblacklist wacom\n" | sudo tee /etc/modprobe.d/blacklist-opentabletdriver.conf
       rm -rf "${OTD_TMPDIR}"
+      # Reload udev rules and immediately unload conflicting drivers so a reboot is not required
+      sudo udevadm control --reload-rules && sudo udevadm trigger
+      sudo modprobe uinput
+      sudo rmmod wacom hid_uclogic 2>/dev/null || true
 
       flatpak --system install -y flathub net.opentabletdriver.OpenTabletDriver
 
@@ -43,7 +49,8 @@ install-opentabletdriver:
       systemctl enable --user --now opentabletdriver.service
     elif [ "${EXIT_CODE}" == 1 ] ; then
       echo "Uninstalling OpenTabletDriver..."
-      sudo rm -f /etc/modprobe.d/blacklist-opentabletdriver.rules /etc/udev/rules.d/71-opentabletdriver.rules
+      sudo rm -f /etc/modprobe.d/blacklist-opentabletdriver.conf /etc/udev/rules.d/71-opentabletdriver.rules
+      sudo udevadm control --reload-rules && sudo udevadm trigger
       flatpak --system remove -y flathub net.opentabletdriver.OpenTabletDriver
     fi
 


hanthor (commented):
## Code Review ✅

**Changes**: 1 file, apps.just (+9/-2).

### Correctness
- ✅ Cleans up old udev rules (90-/99-) before installing new (71-) — prevents rule conflicts from previous OTD installs
- ✅ `udevadm control --reload-rules && udevadm trigger` correctly applies new rules immediately
- ✅ `rmmod wacom hid_uclogic` unloads conflicting drivers without reboot, with `|| true` guard for when not loaded
- ✅ `modprobe uinput` ensures the uinput module is available for OTD

### Bug fix
- ✅ Uninstall path now removes `blacklist-opentabletdriver.conf` instead of `.rules` — was a copy-paste bug

### Closes #340
- ✅ OTD now properly blacklists and unloads conflicting kernel drivers

### Verdict
Solid fix, well-scoped. LGTM.

hanthor (commented):
Clean fix. Key improvements:

**Correct**:
- Cleans stale udev rules (70-, 90-, 99-) before writing new 71- — prevents conflicts
- `udevadm control --reload-rules && udevadm trigger` removes reboot requirement — main UX pain point addressed
- `modprobe uinput` ensures kernel module loaded before driver starts
- `rmmod wacom hid_uclogic 2>/dev/null || true` handles already-unloaded case gracefully
- Fixed uninstall path from `.rules` to `.conf` for modprobe blacklist
- Adds udev reload on uninstall too

**No issues.** Diff is minimal, targeted, and addresses #340 exactly.

castrojo:
## 🤖 Copilot Test Report

**Branch:** `fix/otd-udev-reload-and-uninstall` | **15 commits behind main** | Tested: 2026-05-30T04:51Z

### Test Results

| Test | Result |
|---|---|
| `just check` (syntax validation) | ✅ PASS |
| Branch freshness | ✅ Near head |

### Notes
Adds udevadm reload + modprobe after OpenTabletDriver install/uninstall. Correct fix — udev rules need reload to take effect without a reboot.

ghost:
Hey, I don't want this to come off as I don't appreciate y'all helping me to ensure I can keep doing artwork on ublue based images, and bootc images in general.

But I know Hec and Izzy are currently working on a just script themselves, and I would personally be more comfortable if we just wait for that before merging. This is extremely sensitive for my workflow, so a non AI generated solution would be make me feel better. Especially since Aurora shares this stuff.

Also part of the issue is that I don't believe the script is fully addressing (correct me if im wrong), is the fact that the rules will slowly drift over time, so there needs to be some kind of way for them to be updated for the user at set intervals without having to uninstall and reinstall.

inffy (changes requested):
Lets wait for the proper fix for this, as this considers Aurora too

PR #385, opened 2026-05-31
Fix LUKS UUID parsing on Dakota

dylanmtaylor:
I used Gemini to help debug this issue:

### Root Cause

  The error occurs because the system script  /usr/bin/luks-tpm2-autounlock  (invoked by  ujust toggle-tpm2 ) has a parsing bug.

  It attempts to find the boot LUKS device by searching  /proc/cmdline  strictly for the string  rd.luks.uuid :

    RD_LUKS_UUID="$(xargs -n1 -a /proc/cmdline | grep -F -e "rd.luks.uuid" | cut -d= -f 2 | sed "s/^luks-//" )"

  However, on your machine (and many Fedora Silverblue / Bluefin installations), the encrypted volume is instead specified using the  rd.luks.name  format:

    rd.luks.name=16eb9a62-8dbb-4740-bb40-2185c759d621=root

  Because the script only looks for  rd.luks.uuid , it resolves  RD_LUKS_UUID  as empty, fails the validation check, and exits with:

  │ Could not find LUKS device used to boot system on system mount table. Is your drive encrypted?

## Bug Fixes

* **Bug Fixes**
  * Improved LUKS TPM2 auto-unlock functionality to recognize and properly handle multiple kernel parameter formats for enhanced system compatibility.

### DIFF
--- a/system_files/shared/usr/bin/luks-tpm2-autounlock
+++ b/system_files/shared/usr/bin/luks-tpm2-autounlock
@@ -14,7 +14,7 @@ case "${EXIT_CODE}" in
     exit 0
 esac
 
-RD_LUKS_UUID="$(xargs -n1 -a /proc/cmdline | grep -F -e "rd.luks.uuid" | cut -d = -f 2 | sed "s/^luks-//" )"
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
Tested working for me.

image

renner0e (approved):
I did not test this in it's entirety but this seems to not break anything on Fedora at least and still returns the same uuid

inffy (approved):

Issue #380, opened 2026-05-30
Modernizing the motd


theMimolet:
### What is it?

I'm currently developing a brand motd system made in go.
Its goal is to have a fully translated motd for non-english users and make it extra-configurable 😄
(built-in accent colors in it compatible with gnome too !) 

### What does done look like?

It would result in a binary (shipping method not decided yet) + a common config for all the bluefins ! (except maybe for the DX edition, so it has special developer tips added on it) 

And have it as a full on replacement of the current motd scripts

(and the end goal would also be to share it with Bazzite and Aurora)

### Affected area

brew / CLI tools

### Extra context

# I'll be doing a PR soon-ish myself (I can't self-assign this 😢 ) 

Also a WIP of it on my father's Bluefin :> 
(and yes I also added the dino in the meantime 🦖) 

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/e46e3e5c-ff93-41fa-8b6e-1c2905905c44" />

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/652bcdd2-9cd7-456e-9dc4-bd5f5f47ca20" />

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/021d46e0-ac98-4dc0-8195-9805008ac07a" />

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/67e90a64-8079-461a-9933-b28afefe9533" />

**Common · issue pipeline**

```text
  ✓  triage     needs kind/ + area/ then /approve
  ▶  discussing —
  ·  queued     —
  ·  claimed    theMimolet
  ·  done       —
```

**area:** area/bling, area/brew &nbsp; **priority:** —  
**maintainer:** reach consensus, then comment /approve  
**reporter:** add spec, context, or requirements to the issue body

castrojo:
Beautiful!

castrojo:
This issue has been marked `status/discussing` as part of the factory issue pipeline.

**Next step for a maintainer:** review the proposal, discuss in comments, then comment `/approve` when there is consensus. That moves the issue to `status/queued` where contributors can comment `/claim` to pick it up.

If this needs more definition first, leave a comment with what's missing. If it should be closed or deprioritized, close it directly.

PR #400, opened 2026-06-02
feat: replaces the old motd with umotd

theMimolet:
Hello ! :wave: 

This is the first draft for the [umotd](https://github.com/theMimolet/umotd/) integration in Bluefin. #380 

I removed all the files related to the previous motd and replaced them by a new profile.d and the config file.

The `ublue-dev` tips-preset is temporarily here while I/we figure out a convenient way to deal with them for specific images / scripts. 

Finally, regarding the distribution of it, I still have a few questions : 
- Will it be featured in one of the ublue taps ? Or is [mine](https://github.com/theMimolet/homebrew-tap) just fine ? 
- How will users get it when they first install the system / update from the motd ? (will there be a service that auto-installs it for them ?)

PS: I just saw [this](https://github.com/ublue-os/aurorafin-shared), so it will probably wait as a draft for a bit. I'll do the appropriate changes after it happens.


castrojo:
Let's do this on the image since it needs to be baked in since we use it for alerts and stuff too. Thanks for working on this! +1 from me!

renner0e:
> * Will it be featured in one of the ublue taps ? Or is [mine](https://github.com/theMimolet/homebrew-tap) just fine ?
> 
> * How will users get it when they first install the system / update from the motd ? (will there be a service that auto-installs it for them ?)

> PS: I just saw [this](https://github.com/ublue-os/aurorafin-shared), so it will probably wait as a draft for a bit. I'll do the appropriate changes after it happens.

Like jorge mentioned, this has to be baked into the image and not a good idea to put it in a shared repo, which generally should be a place with not super many experimental changes.

While this is very cool and the proper way to do things™ and does solve the issue of localization, which is generally unpleasant to implement in shell scripts (and the reason why we haven't done this after over 2 years of having the motd). I think it would probably be a good idea to let this cook a little bit more, perhaps this could be tested via brew as an opt-in thing first (and additionally outside of bluefin, others might find it useful as well!). But it sadly is more complex than a shell script, which does lower the amount of people who can maintain it longterm by quite a bit.

castrojo (changes requested):
🚧 **Needs packaging step**: `umotd` is not installed by the image (not in the Containerfile and not in any package list). The new `umotd.sh` profile would run `umotd` as a command that doesn't exist, silently breaking MOTD for all users.

Before this can merge:
1. Add `umotd` to the Containerfile (install via dnf/rpm-ostree/flatpak as appropriate)
2. Confirm `umotd` is available on the correct package source for Bluefin's Fedora base

The rest of the PR looks well-structured — config.json, the profile.d change, removing the old template/env approach. Just needs the install step.

theMimolet:
Well I can do a copr for sure.
But what about Dakota ?

castrojo:
Two blockers before this can land:

1. **Packaging**: `umotd` needs to be installed in the `Containerfile` (e.g., via dnf/rpm-ostree or a brew formula). Without this, the profile.d script runs a command that doesn't exist.

2. **Shared/ boundary**: Several files in this PR touch `system_files/shared/` which is read-only here (the `ublue-os/aurorafin-shared` submodule). Those changes need to go to [`ublue-os/aurorafin-shared`](https://github.com/ublue-os/aurorafin-shared) first — please report upstream manually; agents cannot file PRs in ublue-os repos.

The `system_files/bluefin/` changes look well-structured and can land once the above are addressed.

theMimolet:
Oh well it seems like it's done 😳

PR #349, opened 2026-05-26
add dynamic wallpaper timer

joshyorko:
## Summary

- Add a shared user systemd timer and service for Bluefin dynamic wallpapers.
- Add a GeoClue-backed latitude helper that falls back cleanly when location is unavailable or denied.
- Add a user setup hook so the timer is enabled for users through the common layer.

## Why

This ports the applicable shared-layer work from ublue-os/bluefin#3890 into `projectbluefin/common`, where shared Bluefin workload behavior belongs. That keeps the feature in one layer for downstream Bluefin variants instead of landing it only in `ublue-os/bluefin`.

## Scope note

The approved Bluefin PR also removed the old build-time month rewrite from `ublue-os/bluefin` at `build_files/base/05-override-install.sh`. That file does not exist in `projectbluefin/common`, so this PR cannot remove that line directly. If that rewrite still runs in Bluefin after this common change lands, it should be removed in a small follow-up PR against `ublue-os/bluefin`.

## Testing

- `shellcheck system_files/shared/usr/libexec/bluefin-dynamic-wallpaper system_files/shared/usr/libexec/get-geoclue-latitude system_files/shared/usr/share/ublue-os/user-setup.hooks.d/20-dynamic-wallpaper.sh`
- `bash -n system_files/shared/usr/libexec/bluefin-dynamic-wallpaper system_files/shared/usr/libexec/get-geoclue-latitude system_files/shared/usr/share/ublue-os/user-setup.hooks.d/20-dynamic-wallpaper.sh`
- `git diff --check HEAD~1..HEAD`
- `systemd-analyze verify` with a temporary host-valid `ExecStart` relocation
- Fake `gsettings` test confirmed southern hemisphere maps May to November and updates both light/dark wallpaper URIs
- Fake `gsettings` test confirmed custom dark wallpaper is not overwritten
- Fake `gdbus` tests confirmed valid latitude, AccessDenied exit 2, and `(0.0, 0.0)` rejection

`just check` was also run. It fails on existing formatting drift in untouched Justfiles under `system_files/bluefin/usr/share/ublue-os/just/`; this PR does not touch `Justfile` or `*.just`, so that path-filtered workflow should not run for this change.

Refs ublue-os/bluefin#3890

## Summary by CodeRabbit

* **New Features**
  * Dynamic monthly wallpaper that selects images per hemisphere and updates automatically.
  * Automatic daily wallpaper refresh with safe behavior to preserve custom user-set wallpapers.

* **Chores**
[...]

PR #538, opened 2026-06-08
fix: set summer wallpaper default and run on first login

castrojo:
## Problem

New install ISOs were showing the December winter wallpaper. Root cause: **two issues**:

### 1. Hardcoded December default in gschema override

When wallpapers were first added in PR #34 (December 2025), the gschema override was correctly set to `12-bluefin.xml`. It was never updated after that. PR #349 added the dynamic wallpaper timer but didn't update the static default — so every fresh install still shows December wallpaper as the initial value.

### 2. 1-minute delay before timer fires

The user-setup hook only enabled the timer (`OnBootSec=1min`), meaning there was a visible window where the wrong wallpaper was shown on first login.

## Fix

1. **Update gschema default** to `06-bluefin.xml` for the current summer season
2. **Run `bluefin-dynamic-wallpaper` immediately** in the user-setup hook (in addition to enabling the timer), so the correct season wallpaper is applied the moment the hook runs — no delay, no wrong wallpaper flash

The script already handles geoclue failures gracefully (falls back to northern hemisphere), so running it eagerly is safe.


castrojo:
Closing — the timer self-heals within 1 minute of first boot. Not worth the merge complexity.

PR #569, opened 2026-06-10
chore: simplification audit — remove dead weight, harden edge cases

castrojo:
## Summary

Cleanup from a bang-for-the-buck audit of everything in `common`. Full audit report: [`docs/quality/ARCHITECTURE-REVIEW.md`](docs/quality/ARCHITECTURE-REVIEW.md).

**One item explicitly out of scope:** `lifecycle.yml` move to `projectbluefin/actions` — tracked separately, too large for this PR.

**MOTD consolidation** also out of scope — explicitly deferred.

---

### Deleted

| File | Why |
|---|---|
| `.github/workflows/backfill-pipeline.yml` | One-shot `workflow_dispatch` utility, needed once during migration. Widget logic had already diverged from `lifecycle.yml`'s copy. Will bitrot silently. Move to a local script if needed again. |
| `.github/workflows/skill-drift.yml` | Process convention enforced as a CI gate — explicit violation of AGENTS.md: *"Process conventions are self-enforced by agents. Never implement a process convention as a CI gate."* Skill update discipline stays in the agentic review loop. |
| `system_files/bluefin/etc/profile.d/open.sh` | Entire file was `alias open="xdg-open &>/dev/null"` — one line does not warrant a `profile.d` file. Alias moved to `bling.sh`. |

### Modified

**`bling.sh`** — Remove dead atuin integration
- `ATUIN_INIT_FLAGS` variable was set but referenced only by a permanently-commented-out `eval` line — disabled-but-present dead code shipped in every user's shell session
- Add `open → xdg-open` alias (consolidated from deleted `open.sh`)

**`libsetup.sh`** — Harden two silent failure modes
- Add `flock -x` guard around the check/write cycle: concurrent first-boot setup scripts (user-setup + privileged-setup) could both read the JSON before either writes back, causing duplicate execution of setup steps
- Add JSON validation with safe reset: malformed `setup_versioning.json` previously caused `jq` to exit non-zero, leaving the temp file uncleaned and silently skipping setup as if it had already run
- Add `trap 'rm -f "${tmp}"' EXIT` on temp file in the subshell
- All 9 existing bats tests pass unchanged ✅

**`sync-codeowners.yml`** — Add dry-run mode
- New `dry_run: boolean` input on `workflow_dispatch`: shows exactly what would change without making any API calls
- This workflow makes irreversible `PUT /collaborators` and `DELETE /collaborators` calls across 4 downstream repos — a typo in the CODEOWNERS triagers block previously had no preview option
- Add `set -euo pipefail` to both run steps (was missing)

**`.gitignore`** — Add `.coverage`, `.wrangler/`, `knuckle/`, `__pycache__/`, `*.pyc`, `.worktrees/` which were accumulating as untracked noise

---

## Verification

```
just check       ✅ all Justfiles valid
pre-commit       ✅ json/yaml/toml/actionlint/SHA-pinning all passed
just test        ✅ 20 pytest + 45 bats = 65 tests passed
shellcheck       ✅ bling.sh + libsetup.sh clean
```

## Summary by CodeRabbit

* **Chores**
  * Removed backfill-pipeline and skill-drift GitHub Actions workflows.
  * Updated .gitignore with additional ignore patterns.

* **New Features**
  * Added dry-run mode to sync-codeowners workflow for safe preview of changes.

* **Bug Fixes**
  * Enhanced version-script with file locking to prevent concurrent setup race conditions.
  * Improved JSON validation and error handling in setup services.

* **Refactor**
  * Restructured shell initialization and updated command aliases.


castrojo:
## Review — #569 chore: simplification audit — remove dead weight, harden edge cases

**Verdict: GO** (critical fixed inline)

### Fixed in follow-up commit (1367a32)
- `libsetup.sh` — bash opens `200>"${lock_file}"` before the subshell body executes. On first boot, `$HOME/.local/share/ublue/` doesn't exist yet, so the redirect fails before `flock` runs, and `version-script` exits non-zero for every caller — all setup scripts skip silently. Fix: `mkdir -p` the parent directory before the subshell+flock block. Verified: `bats tests/test_libsetup.bats  9/9 ✅`.

### Worth noting (no action required)
- `bling.sh` — consolidating the `open → xdg-open` alias from `profile.d/open.sh` into `bling.sh` changes its scope: was login-shell-wide for all users; now opt-in (only users who source bling get it). Intentional or incidental? Fine either way if intentional — just worth being explicit.
- `sync-codeowners.yml` — `set -euo pipefail` addition is safe for the specific grep-in-if pattern used. Recommend a smoke run with `dry_run=true` before the next real triager change.
- `libsetup.sh:21` — malformed-JSON silent reset is acceptable for version tracking but consider a `logger -t ublue-setup` call so resets are visible in the journal.

### Strengths
- `skill-drift.yml` deletion is the right call — process conventions in CI gates is an explicit AGENTS.md violation.
- `backfill-pipeline.yml` deletion is clean — one-shot utility with no further purpose.
- `sync-codeowners.yml` dry-run mode is a real safety win for an irreversible multi-repo operation.
- libsetup.sh flock + JSON validation + trap is a meaningful hardening, correct once the mkdir ordering is fixed.

CI: all checks green (tests:pass label set).

PR #573, opened 2026-06-10
test(quality): bats tests for rechunker-group-fix, ublue-bling-fastfetch, changelog.just + coverage gate

castrojo:
## Summary

Closes #571, #572, #561. Three quality issues in one PR.

---

### rechunker-group-fix
- Add `GSHADOW_FILE` / `GROUP_FILE` env-var overrides for test isolation
- Comment flags this as a workaround (upstream: systemd/systemd#30852) and marks both the script and the overrides for removal when upstream resolves it

### changelog.just
- Apply LTS repo-selection logic from PR #542/#543 (which are still pending review)
- `^lts` tag → `projectbluefin/bluefin-lts`, all others → `projectbluefin/bluefin`
- Fixes the hardcoded `ublue-os/bluefin` org reference

### New test files (3)

| File | Tests | Covers |
|---|---|---|
| `tests/test_rechunker_group_fix.bats` | 7 | append, duplicate detection, multi-group, empty file, format |
| `tests/test_bling_fastfetch.bats` | 17 | all 9 accent colors, default fallback, dconf/gsettings chain, quote stripping, FASTFETCH_FORCE_THEME precedence |
| `tests/test_changelog.bats` | 12 | repo selection (lts/stable/gts/latest), URL construction, exit behaviour |

### unit-tests.yml
- Add `--cov-fail-under=80` to pytest (currently at 100%, gate is forward-looking)
- Wire 3 new bats steps for the new test files

---

## Verification

```
just check                    ✅ all Justfiles valid
pre-commit run --all-files    ✅ json/yaml/toml/actionlint/SHA-pinning passed
bats test_rechunker_group_fix 7/7  ✅
bats test_bling_fastfetch     17/17 ✅
bats test_changelog           12/12 ✅
pytest --cov --cov-fail-under=80   20/20 passed, 100% coverage ✅
```

---

**Note on changelog.just overlap:** PRs #542 and #543 also touch `changelog.just`. If either merges first, this PR will need a rebase to resolve the conflict — the logic is identical so the merge is trivial.

PR #580, opened 2026-06-10
fix(test): add IMAGE_INFO_FILE override to changelog.just, fix changelog bats in CI

castrojo:
## Summary

Fixes the CI failure on PR #573. Tests 1, 2, 3, 7, 9 in `test_changelog.bats` were failing.

## Root Cause

`changelog.just` reads the image tag via:
```bash
TAG="$(jq -r '.["image-tag"]' < /usr/share/ublue-os/image-info.json)"
```
In the GitHub Actions CI environment `/usr/share/ublue-os/image-info.json` does not exist. The bash shell fails the stdin redirect **before** jq is invoked, leaving `TAG` empty. With `TAG` empty:
- LTS branch is never taken → repo always defaults to `projectbluefin/bluefin`
- Tag-specific release lookup is skipped → always falls back to `/releases/latest`

Tests 4, 5, 6, 8 passed because they only asserted `projectbluefin/bluefin` (the fallback) was called — which it was.

## Fix

**`changelog.just`** — add `IMAGE_INFO_FILE` env-var override, matching the established pattern used by `ublue-privileged-setup`, `luks-tpm2-autounlock`, etc.:
```bash
IMAGE_INFO_FILE="${IMAGE_INFO_FILE:-/usr/share/ublue-os/image-info.json}"
TAG="$(jq -r '.["image-tag"]' < "${IMAGE_INFO_FILE}")"
```

**`tests/test_changelog.bats`** — create a temp `image-info.json` in `WORKDIR` and export `IMAGE_INFO_FILE` in `setup()`, with a comment explaining why this is required.

## Verification

```
bats tests/test_changelog.bats   12/12 ✅
just check                         ✅
pre-commit run --all-files         ✅
```

This PR is intended to be squashed into PR #573 or merged first so #573's CI goes green.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -51,7 +51,7 @@ jobs:
         run: find system_files -name '*.sh' -print0 | xargs -0 shellcheck -e SC1091
 
       - name: Run pytest (hooks.py)
-        run: python3 -m pytest tests/test_hooks.py -v --cov=tests --cov-report=term-missing
+        run: python3 -m pytest tests/test_hooks.py -v --cov=tests --cov-report=term-missing --cov-fail-under=80
 
       - name: Run bats (libsetup.sh)
         run: bats tests/test_libsetup.bats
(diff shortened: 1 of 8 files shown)


castrojo:
Fix applied directly to PR #573's branch (commit e47032e). Closing this PR as redundant.

PR #669, opened 2026-06-13
fix(brew): trust ublue-os/tap and ublue-os/experimental-tap for Homebrew 6.0

castrojo:
Closing as superseded by PR #672, which includes all changes from this PR in its second commit (brew tap trust fix). Will land as part of #672.


hanthor (dismissed):

hanthor (commented):
## Adversarial review note

### [BLOCKER] Superseded by common#672
common#672 includes all changes from this PR in its second commit. Both are open. This PR should close once #672 is ready to merge.

### [MEDIUM] Repeated tap warning on devmode toggle
`system.just:138-143` — `brew tap --trust` now warns "Already tapped!" every time devmode is toggled. Previously `2>/dev/null || true` suppressed this. The `|| true` removal is correct (errors should surface), but the "already tapped" case is not an error.

**Fix:** Add `2>/dev/null` (not `|| true`) for the pre-existing success case: `brew tap --trust ublue-os/tap 2>/dev/null`. Genuine failures still surface.

Issue #559, opened 2026-06-10
[quality] ublue-rollback-helper has zero behavior tests


clubanderson:
## Finding

`system_files/bluefin/usr/bin/ublue-rollback-helper` is a user-facing image rebase tool with complex branching logic:

- LTS mode detection via image tag
- Tag filtering with `skopeo list-tags` and regex
- Channel selection (stable/stable-daily/latest vs lts/lts-hwe)
- Image variant mapping (bluefin/bluefin-dx/nvidia variants)
- Conditional rebase via `bootc switch` or `rpm-ostree rebase`

The script is ~80 lines with interactive `gum` prompts, but the **pure logic** (LTS detection, tag filtering, image/channel mapping, rebase target construction) can be extracted into testable functions.

**Risk:** Incorrect rebase target construction could strand users on wrong image channels or break updates entirely.

## Recommendation

1. Extract pure functions (`list_tags`, LTS detection, channel-to-image mapping, rebase target assembly) into a sourceable helper
2. Write bats tests covering:
   - LTS mode detection from image tag
   - Tag filtering regex for each channel type
   - HWE vs non-HWE tag separation
   - Rebase target string construction for all variants
3. Add to `unit-tests.yml` workflow

## Priority
- Impact: **high** (user-facing, data-loss risk if wrong rebase target)
- Effort: medium (requires refactoring interactive script into testable pieces)

---
*Filed by quality agent (hold-gated mode)*

**Common · issue pipeline**

```text
  ▶  triage     needs kind/ + area/ then /approve
  ·  discussing —
  ·  queued     —
  ·  claimed    —
  ·  done       —
```

**area:** — &nbsp; **priority:** —  
**maintainer:** set kind/ + area/ labels, then comment /approve  
**reporter:** answer questions; add repro steps if a bug

castrojo:
Closing as duplicate of #564 (ublue-rollback-helper complex branching logic untested), which contains the more detailed analysis and test recommendations. All context from this issue is captured there.

PR #675, opened 2026-06-13
feat(nvidia): enable nvidia-cdi-refresh systemd preset for CDI GPU passthrough

castrojo:
Closing as redundant with upstream RPM.

The `nvidia-container-toolkit-base` RPM (which Bluefin NVIDIA images already ship) includes `90-nvidia-container-toolkit.preset` in its `%package base` section, which enables the same `nvidia-cdi-refresh.path` and `nvidia-cdi-refresh.service` units that this PR targets.

Verified via: https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/packaging/rpm/SPECS/nvidia-container-toolkit.spec

Adding a custom `80-nvidia-container-toolkit.preset` would shadow the upstream preset, masking future upstream changes with no functional benefit.

hanthor review confirmed the gap — closing rather than merging.

### DIFF
--- a/system_files/nvidia/usr/lib/systemd/system-preset/80-nvidia-container-toolkit.preset
+++ b/system_files/nvidia/usr/lib/systemd/system-preset/80-nvidia-container-toolkit.preset
@@ -1,0 +1,10 @@
+# Enable CDI spec auto-generation so 'podman run --device nvidia.com/gpu=all' works
+# out of the box on nvidia image variants.
+#
+# nvidia-cdi-refresh.path watches /lib/modules/*/modules.dep and /usr/bin/nvidia-ctk
+# and re-triggers nvidia-cdi-refresh.service whenever the driver or toolkit changes.
+# The service runs 'nvidia-ctk cdi generate' and writes /var/run/cdi/nvidia.yaml.
+#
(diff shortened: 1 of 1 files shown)


hanthor (commented):
## Adversarial review — optional improvement

### [HIGH] Redundant with upstream NVIDIA preset
`system_files/nvidia/usr/lib/systemd/system-preset/80-nvidia-container-toolkit.preset` is functionally identical to NVIDIA upstream `90-nvidia-container-toolkit.preset` shipped in the `nvidia-container-toolkit-base` RPM. Both enable the same units. Systemd handles duplicate enables gracefully, but the duplication masks future upstream changes and adds maintenance.

**Fix:** Verify the upstream RPM ships the preset. If so, drop this custom file. If the RPM does NOT include the preset (e.g., `-base` variant excludes it), document the gap explicitly so future maintainers know why the custom preset is needed.

### [MEDIUM] Cross-PR dependency on bluefin#559
This preset enables `nvidia-cdi-refresh.*` units that are installed by bluefin#559. If this PR merges before #559, the preset enables non-existent units (systemd ignores missing preset targets, so it is safe but silent). The dependency is mentioned in the PR body but not enforced by CI.

**Note:** No action needed if merge order is coordinated — just making the dependency explicit.

PR #708, opened 2026-06-14
fix: set summer wallpaper default and run on first login

castrojo:
The gschema override was hardcoded to `12-bluefin.xml` (December) from when wallpapers were first added in December 2025. This caused all new ISO installs to show the winter wallpaper regardless of season.

## Changes

1. **Update gschema default** to `06-bluefin.xml` for the current summer season (`zz0-bluefin-modifications.gschema.override`)
2. **Run dynamic wallpaper immediately** in the user-setup hook so the correct season is applied on first login, not after the 1-minute timer delay (`20-dynamic-wallpaper.sh`)

## Why both?

The gschema fix covers the state before first login (GDM, initial GNOME session). The hook fix covers the edge case where a user installs in June but the timer hasn't fired yet — the wallpaper now updates immediately on first login rather than showing whatever month the gschema hardcoded.

Note: The gschema will need to be updated each season going forward, or better yet replaced with a dynamic default that reads the current month. That's a follow-up; this is the minimum fix for the current summer season.

## Release Notes

* **Bug Fixes**
  * Updated the default desktop background image configuration to ensure proper and consistent rendering in both light and dark display modes.

* **New Features**
  * Dynamic wallpaper functionality now automatically initializes when your user session starts, providing an improved and consistent visual experience from your first login onwards.


castrojo:
Changes already landed directly on main via cherry-pick. Closing stale PR.

PR #732, opened 2026-06-20
fix(ci): dispatch common-updated to dakota after manifest publish

castrojo:
## What

After the multi-arch manifest is signed and published, dispatch a `repository_dispatch` event (`common-updated`) to `projectbluefin/dakota` using the mergeraptor token.

## Why

dakota was picking up common updates once per day at 06:00 UTC via a cron. bluefin and bluefin-lts pick them up within minutes via Renovate. This makes dakota match.

## Companion PR

[projectbluefin/dakota#942](https://github.com/projectbluefin/dakota/pull/942) — adds the `repository_dispatch` trigger and CI-gated `testing`-branch flow to dakota's `track-bst-sources.yml`. Both PRs need to land for the event-driven path to activate. Dakota's daily cron remains as fallback.

## Requirements

`MERGERAPTOR_APP_ID` and `MERGERAPTOR_PRIVATE_KEY` secrets must be configured in this repo (common). The mergeraptor app must be installed on `projectbluefin/dakota`.


castrojo:
Closing — wrong approach. Cross-repo dispatch requires secrets that shouldn't be needed. dakota will self-poll instead.

PR #767, opened 2026-06-22
fix(flatpak): refresh appstream on every boot instead of firstboot only

castrojo:
Remove the firstboot-only condition from flatpak-appstream-firstboot.service so appstream metadata is refreshed on every boot after network is up, before the graphical session starts.

Bazaar calls flatpak_installation_update_appstream_full_sync internally, but when running inside the Flatpak sandbox this can fail, leaving stale cached data. Pre-refreshing as root before the user session starts ensures current metadata is always available on launch.

Changes:
- Remove ConditionPathExists=!/var/lib/flatpak/.appstream-refreshed (firstboot restriction)
- Remove ExecStartPost marker file touch
- Change WantedBy from multi-user.target to graphical.target (runs before user session, after network)

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified Flatpak appstream metadata refresh behavior to occur more regularly instead of only during initial system boot.
  * Updated service configuration to ensure appstream management runs properly in graphical environments.

### DIFF
--- a/system_files/shared/usr/lib/systemd/system-preset/02-flatpak-appstream-firstboot.preset
+++ b/system_files/shared/usr/lib/systemd/system-preset/02-flatpak-appstream-firstboot.preset
@@ -1,1 +1,0 @@
-enable flatpak-appstream-firstboot.service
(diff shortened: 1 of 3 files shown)


hanthor (dismissed):
LGTM. Reviewed 2026-06-24 — CI green, standards clean, changes well-scoped.

PR #768, opened 2026-06-22
feat(uupd): smart AC-aware update scheduling

castrojo:
## What

Gate all uupd runs on AC power, update every 6 hours, and trigger immediately when a laptop is plugged in.

## Why

Updates were running on battery, draining charge unexpectedly. The existing uupd battery check (20% threshold) is too permissive — a bootc image pull at 50% battery on an untethered laptop is bad UX. macOS and Windows both defer updates entirely to AC.

The 6h timer alone is not enough for laptops that are on battery during all scheduled windows. The AC-connect trigger closes that gap.

## Changes

- uupd.timer: runs at 00:00, 06:00, 12:00, 18:00 with Persistent=true (sleeping machines catch up on next wake while on AC)
- uupd.service.d/10-bluefin.conf: ConditionACPower=true — hard gate, no updates on battery
- uupd-on-ac.service: oneshot triggered by udev, waits 60s for network to settle, then starts uupd.service
- udev/rules.d/99-uupd-on-ac.rules: fires uupd-on-ac.service on ACTION==change for Mains power supply coming online

## Behavior matrix

| Scenario | Result |
|---|---|
| Desktop (always AC) | Updates every 6h |
| Laptop on AC at timer fire | Updates every 6h |
| Laptop on battery at timer fire | Skips cleanly, no drain |
| Laptop just plugged in | Updates after 60s network settle |
| Machine wakes from sleep on AC | Persistent=true catches up |
| Metered WiFi | Blocked by existing uupd hw-check |
| Power-saver mode | Blocked by existing uupd hw-check |

## Summary by CodeRabbit

* **New Features**
  * Added automatic update scheduling with a new timer that runs several times a day and persists across reboots.
  * Updates can now start automatically when AC power is connected, after a short delay and once networking is ready.

* **Bug Fixes**
  * Restricted update activity to AC power to help avoid running during battery use.
  * Added rate limiting to prevent repeated rapid starts.

### DIFF
--- a/system_files/shared/usr/lib/systemd/system/uupd-on-ac.service
+++ b/system_files/shared/usr/lib/systemd/system/uupd-on-ac.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Trigger system updates on AC power connect
+After=network-online.target
+Wants=network-online.target
+ConditionACPower=true
+
+[Service]
+Type=oneshot
+# Wait for network to settle after AC connect before triggering updates
+ExecStartPre=/bin/sleep 60
+ExecStart=/usr/bin/systemctl start uupd.service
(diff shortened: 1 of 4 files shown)


hanthor (approved):
LGTM. Reviewed 2026-06-24 — CI green, standards clean, changes well-scoped.

PR #807, opened 2026-07-01
feat(telemetry): implement weekly countme client for dakota

castrojo:
## Summary

Add a lightweight, weekly-salted, privacy-preserving telemetry client for Dakota using systemd `DynamicUser` and `curl`, matching the Cloudflare Worker server implementation.

## Changes

- `system_files/shared/usr/libexec/dakota-countme` — shell script that generates a salted weekly HMAC-SHA256 hash from `/etc/machine-id` and POSTs it to the countme endpoint
- `system_files/shared/usr/lib/systemd/system/dakota-countme.service` — one-shot service running under a transient `DynamicUser` with no persistent state
- `system_files/shared/usr/lib/systemd/system/dakota-countme.timer` — weekly timer with a ±12h randomized delay to prevent thundering-herd
- `system_files/shared/usr/lib/systemd/system-preset/03-dakota-countme.preset` — enables the timer by default

## Privacy design

- No persistent identifier is transmitted — the weekly hash is derived from `machine-id` salted with the ISO week number, so the same value cannot be correlated across weeks
- The service runs as a transient `DynamicUser` (no home dir, no persistent UID)
- Only a weekly count ping is sent; no user data, hostname, or IP is stored server-side beyond what Cloudflare Workers logs transiently

## Testing

Verified files are present and the service/timer unit syntax is valid.

## Summary by CodeRabbit

* **New Features**
  * Added weekly “Dakota Count Me” telemetry reporting via a new scheduled system service/timer.
  * Bazaar previews now auto-generate PNG banner assets from JXL when branding files are present.

* **Bug Fixes**
  * Improved Bazaar service behavior by switching to a long-running service mode.
  * Updated OCI reference validation to ignore non-production tag patterns (e.g., end-to-end PR tags).

* **Documentation**
  * Reworked lifecycle guidance to a GitOps “Branch-as-State” model and retired legacy label/FSM workflows across skills and governance docs.
  * Updated templates and contributor guidance to match the new branch/PR flow.
  * Refreshed Bazaar/factory preview and verification instructions.

* **Tests**
  * Added coverage for skipping e2e-pr tags and made changelog tests more isolated.


hanthor (changes requested):
This PR currently has merge conflicts with the base branch and can't be merged as-is. Could you rebase / resolve the conflicts? Happy to re-review once it's mergeable — the change itself looks reasonable.

castrojo:
@hanthor rebased onto main, resolved the merge conflicts, and addressed the outstanding feedback:

- Added `Wants=network-online.target` / `After=network-online.target` to `dakota-countme.service`.
- Updated `docs/skills/image-registry.md` to use the actual `/metalink` endpoint.
- Dropped the unrelated `game-devices-udev` checksum refresh that was causing the rebase conflict.

CI is running now; shellcheck + systemd-analyze verify + `just check` + `pre-commit` passed locally. Ready for another look.

ahmedadan (dismissed):

hanthor (changes requested):
The implementation is clean — `StateDirectory` + `DynamicUser`, weekly throttle, `RandomizedDelaySec=12h` to spread load, `lastrun` only written on a successful ping so a flaky network retries. My concerns are about policy and dependencies rather than the shell.

### 1. No opt-out (blocking)

This moves Dakota's telemetry from Fedora's metalink to an endpoint we operate (`countme.projectbluefin.io`), and there is no way for a user to turn it off. Fedora's countme is disableable (`countme=false` in dnf config); ours should have an equivalent. Something as simple as:

```bash
[ -e /etc/dakota-countme/disabled ] && exit 0
```

plus a `ConditionPathExists=!/etc/dakota-countme/disabled` on the unit, and a line in the docs saying how to opt out. Given we're now the ones collecting it, having a documented off switch matters more than it did when we were just riding Fedora's.

While you're there — the docs change says "anonymously," and the request does look unlinkable (no machine ID, no persistent token). Worth stating explicitly in the skill doc what the Worker does and doesn't log, particularly whether source IPs are retained, so that claim is backed by something.

### 2. Confirm `jq` is in the Dakota image

The script is `set -euo pipefail` and calls `jq` three times. If `jq` isn't present the unit fails loudly every week rather than degrading. Dakota is GNOME OS-based, so I don't want to assume it's there the way I would on the Fedora images — please confirm, or gate on `command -v jq` and exit 0.

Same question for `curl`, though I'd expect that one to be present.

### 3. Minor

- `ConditionPathExists=/usr/share/ublue-os/image-info.json` is on both the timer and the service, and the script re-checks it. Harmless belt-and-braces, but the timer condition means a non-Dakota image doesn't even arm the timer, which makes the `image-name != dakota` early-exit in the script the only load-bearing gate on shared images. That's fine — just noting the three checks aren't redundant in the way they look.
- `OS_VERSION` is parsed but only used in the `User-Agent`, not in the query string. Intentional?
- The bucket boundaries (1 / 2-4 / 5-24 / >24 weeks) match Fedora's scheme. Worth a comment linking to Fedora's countme spec so the magic numbers are traceable.

---

Danathar:
@castrojo is this draft superseded by your #1112? That one ships the same weekly countme client for every image via `system_files/shared` (and dakota#1561 switches Dakota to it), so a Dakota-specific client here looks like the older approach. If so, closing this would also let #956 point at #1112 instead.

castrojo:
Superseded by #1154.

PR #815, opened 2026-07-07
feat(motd): add uwelcome + update umotd

theMimolet:
Good day :wave: 

## PR pipeline

```
opened ──▶ review ──▶ approved ──▶ merged
                    [lgtm]      auto-merge
                                when CI green
```

> Add `do-not-merge` at any time to block automation.
> `/approve` or `lgtm` from a maintainer triggers merge queue.

## What does this change?

It replaces the old unified umotd config file by the two new ones for [umotd](https://github.com/theMimolet/umotd) and [uwelcome](https://github.com/theMimolet/uwelcome). 

## Why?

To update them and make them work together.
(also so you can use umotd in other programs as it only prints translated messages)

## Checklist

- [x] `just check` passes
- [ ] `pre-commit run --all-files` passes
- [x] PR title follows Conventional Commits (`fix:`, `feat:`, `chore:`, etc.)

## Summary by CodeRabbit

* **New Features**
  * Added support for additional desktop/profile tags, including Bluefin and GNOME.
  * Updated the welcome/MOTD experience to use `uwelcome toggle`, including updated greeting structure and MOTD command grouping.
* **Bug Fixes**
  * Refreshed welcome and MOTD behavior to improve consistency of the displayed greeting/prompt.
* **Chores / Security**
  * Updated the container build to compile and ship `uwelcome` alongside `umotd`, and refreshed bundled udev asset verification.


castrojo (commented):
I don’t see a ship/install path for uwelcome in this repo. The new config switches the command to , but the image only appears to install/ship                                                                               
  [38;5;104;1m󱍢 Welcome to[m[38;5;104;1m Bluefin[m                                                        
                                                                              
  [Ci] [38;5;104;1m ghcr.io/projectbluefin/dakota:latest [m                                 
                                                                              
   >_ Command                           │ Description                         
  ──────────────────────────────────────┼─────────────────────────────────────
   [38;5;104;1m ujust --choose [m                     │ List all available commands         
   [38;5;104;1m umotd toggle [m                       │ Toggle this banner on/off           
   [38;5;104;1m ujust bluefin-cli [m                  │ Enable terminal bling               
   [38;5;104;1m brew help [m                          │ Manage command line packages        
                                                                              
  [38;5;104;1mSupport indie game preservation and OSS developers![m — ]8;id=1207160287;https://patreon.com/hitsave[1mJoin Hit Save!'s[m]8;;      
  ]8;id=1207160287;https://patreon.com/hitsave[1mPatreon[m]8;; [38;5;104;4m]8;id=1207160287;https://patreon.com/hitsavehttps://patreon.com/hitsave]8;;[m                                         
                                                                              
  • ]8;id=3541281617;https://issues.projectbluefin.io/[1mReport an issue[m]8;; [38;5;104;4m]8;id=3541281617;https://issues.projectbluefin.io/https://issues.projectbluefin.io/]8;;[m                         
  • ]8;id=3766875054;https://ask.projectbluefin.io/[1mAsk Bluefin[m]8;; [38;5;104;4m]8;id=3766875054;https://ask.projectbluefin.io/https://ask.projectbluefin.io/]8;;[m                                
  • ]8;id=772340340;https://docs.projectbluefin.io/[1mDocumentation[m]8;; [38;5;104;4m]8;id=772340340;https://docs.projectbluefin.io/https://docs.projectbluefin.io/]8;;[m                              today. Please either wire up uwelcome in the image or keep the config aligned with the actually shipped binary so the new path is functional.

castrojo (commented):
I don’t see a ship/install path for uwelcome in this repo. The new config switches the command to `uwelcome toggle`, but the image only appears to install/ship `umotd` today. Please either wire up uwelcome in the image or keep the config aligned with the actually shipped binary so the new path is functional.

theMimolet:
tbf I was expecting you to do your magic
but I can try something out

castrojo:
## Vanguard Lab Strike Report

| Field | Value |
|---|---|
| Workflow | `common-815-feat-motd-add-uwelcome-update-l7vdh` (link: `http://192.168.1.102:32746/workflows/argo/common-815-feat-motd-add-uwelcome-update-l7vdh`) |
| Phase | `Failed` |
| Duration | `32m31s` |
| Variants tested | bluefin:testing |
| Suites run | smoke |

## Results
```
Workflow phase: Failed
Failed nodes: 6 / 15 total (3 succeeded, 6 skipped)

assert-cd node:
  phase: Failed
  message: main: Error (exit code 1)

Relevant log/script excerpt:
  TAG="{{inputs.parameters.containerdisk-tag}}"
  CD_IMAGE="192.168.1.102:30500/bluefin-containerdisk:${TAG}"
  DIGEST=$(skopeo inspect --tls-verify=false --format '{{.Digest}}' \
    "docker://${CD_IMAGE}" 2>/dev/null || echo "")
  if [[ -n "${DIGEST}" ]]; then
    echo "containerdisk ${CD_IMAGE} ready: ${DIGEST}"
  else
    echo "ERROR: containerdisk ${CD_IMAGE} not found in Zot" >&2
    echo "The digest-watch CronWorkflow builds it when GHCR changes." >&2
    echo "Check: argo list -n argo --field-selector metadata.name=~build-cd-sync" >&2
    exit 1
  fi

Outcome:
  - The smoke test lane failed before provisioning a VM.
  - The common suite was skipped because the submitted workflow was configured with suites=smoke only.
  - No motd/uwelcome behavior was exercised.
```

## Evidence collected via
- [x] `argo_get_workflow` — workflow phase and node summary
- [x] `argo_logs_workflow` — cleanup-orphan-vms log (test pod logs already recycled)
- [ ] `kubernetes-mcp-resources_list` — VMI state confirmed Ready (n/a: container-only pipeline)
- [x] VM/orphan check — zero orphaned VMs after teardown

## Merge-queue eligibility
- Argo workflow success: no
- No regressions vs base image: no / not determined
- PR mergeable (no conflicts, approvals present): no

## Blockers / issues filed
- The existing Argo run for SHA `18ef56fc3d94` failed at the `assert-cd` step because `bluefin-containerdisk:testing` was not present in the cluster-local Zot registry.
- Only the `smoke` suite was submitted; the requested `common` suite was not run.
- PR has no human approvals (`reviewDecision` empty, label `pr/needs-review`) and `mergeStateStatus` is `UNKNOWN`.

theMimolet:
It should be good now :+1:

hanthor (commented):
Holding a neutral review — the build wiring looks coherent but I can't confirm the greeting path end-to-end from the diff alone, and there are a couple of concrete gaps worth resolving first:

1. **`uwelcome` is installed but nothing in the image invokes it.** The config moves from `etc/umotd/config.json` to `etc/uwelcome/config.json` and the new binary lands at `/usr/bin/uwelcome`, but on this branch both greeting entrypoints — `fish_greeting.fish` (`function fish_greeting; umotd; end`) and `profile.d/umotd.sh` — still call `umotd`, not `uwelcome`. If the new `umotd` is now "only prints translated messages" (per the PR description) and `uwelcome` is the front-end that reads `etc/uwelcome/config.json`, then the greeting needs to switch to `uwelcome`; otherwise the interactive welcome (commands/links) silently regresses. If instead `umotd` still delegates internally, please confirm it reads the new `etc/uwelcome/` path.

2. **`default.just` toggle mismatch.** `umotd toggle` in `default.just` is unchanged, but the new config's toggle command is `uwelcome toggle`. These should agree.

3. **Supply chain:** `uwelcome` is cloned from the personal `github.com/themimolet/uwelcome` rather than an org-owned repo (unlike `umotd` from `projectbluefin/motd`). Intentional, or should it be migrated under the org before it ships in the image?

Minor: the Containerfile carries a lot of unrelated 2→4-space reindentation churn (the gdu checksum block, etc.). The values are unchanged — CodeRabbit's "refreshed SHA-256" note is a misread of the whitespace — but the noise makes the diff harder to audit; consider splitting it out. CI is green. Happy to approve once the greeting wiring (item 1) is confirmed.

theMimolet:
For No. 1 I completely forgot.
For No. 2 I thought it was already the case.
For No. 3 I don't have permission to create repos, so I'm doing what I can 🥲

theMimolet:
Ok, might need some more rechecks

Here's what I did that I don't know if you agree with : 
- I added a line in the profile.d & fish_greeting scripts to remove no-show-user-motd as it's being deprecated and would no longer be used - it will be a line that would stay for not that long since it would just be a transitional thing
- I completely removed the motd toggle command from the ujust default file because people would generally just see the command shown in the banner itself and just do it

Other notes: 
- The CodeRabbit's "refreshed SHA-256" note is normal, there has been an update to the docker image used for building umotd, so I updated it for uwelcome as well
- Regarding the indentation issues, the Container tools extension just formats whenever I save and all, so I changed the spaces for now, even though it's not the same either - which formatter do you use on your end ? 
-  And finally regarding the supply chain stuff, I would feel more comfortable if the repos were transferred directly rather than forked - I'll ask you about it on Discord

hanthor (changes requested):
The greeting wiring is now present, but the migration currently deletes ~/.config/no-show-user-motd on every shell start in both Bash and Fish. That silently re-enables the banner for every user who opted out. uwelcome stores its replacement marker at ~/.config/uwelcome/disabled, so please migrate the legacy marker to that path once instead of erasing the preference, and restore executable tests for the migration and opt-out behavior rather than leaving the relevant assertions commented out.

theMimolet:
yeah my bad, i don't know why I didn't think about it
all good now :+1:

hanthor (changes requested):
The uwelcome migration is a good direction, but there are two things to fix before this ships.

### 1. The fish migration will error on every shell start

`system_files/shared/usr/share/fish/vendor_conf.d/fish_greeting.fish`:

```fish
if test -e ~/.config/no-show-user-motd
    mv ~/.config/no-show-user-motd ~/.config/uwelcome/disabled
end
```

Two problems, and they compound:

- There's no `2>/dev/null` here, unlike the profile.d version.
- `~/.config/uwelcome/` is not created anywhere in this PR. If it doesn't exist, `mv` fails.

So for any user who had opted out via `no-show-user-motd`, the `mv` fails, fish prints an `mv` error, and — because the source file is never removed — it fails again on *every* new fish shell, forever. The profile.d path has the same missing-directory bug; it just hides the error instead of showing it, which means the opt-out silently never migrates.

Both paths need the directory created first, e.g.:

```fish
if test -e ~/.config/no-show-user-motd
    mkdir -p ~/.config/uwelcome
    mv ~/.config/no-show-user-motd ~/.config/uwelcome/disabled 2>/dev/null
end
```

and the equivalent in `uwelcome.sh`. Worth confirming against uwelcome upstream that `~/.config/uwelcome/disabled` is actually the path it checks, since the whole migration hinges on it.

### 2. Two tests were commented out rather than fixed

In `tests/test_motd_integration.bats`:

```bash
# @test "uwelcome.sh: does not contain no-show-user-motd check" {
# @test "fish_greeting: does not check no-show-user-motd (opt-out delegated to uwelcome)" {
```

These were live assertions in `test_umotd_integration.bats` and are now dead. I understand why — the migration shim *does* reference `no-show-user-motd`, so they'd fail as written. But leaving them commented out means the file's own header comment ("Neither profile.d nor fish_greeting should check for that file or gate the uwelcome invocation") no longer describes the code, and nothing enforces the shim's shape.

Please either delete them and update the header comment to describe the migration as it actually is, or rewrite them to assert what you now want — e.g. that the reference appears *only* inside the migration block and that `uwelcome` is invoked unconditionally afterward. A commented-out test is the one outcome that leaves no record of which it was.

### Minor

- `tests/test_umotd_integration.bats` also dropped the three `toggle-user-motd` recipe tests along with the recipe itself. That's consistent (the recipe moved to `uwelcome toggle` in the config JSON), but users with muscle memory for `ujust toggle-user-motd` now get "recipe not found" with no pointer. Consider keeping a stub recipe that execs `uwelcome toggle`.
- The `Containerfile` reindentation is a large share of the diff and makes the substantive uwelcome-build-stage addition harder to spot. Not worth re-rolling for, just noting for next time.

---

system_files/shared/usr/share/fish/vendor_conf.d/fish_greeting.fish
```diff
@@ -1,3 +1,8 @@
 function fish_greeting
-    umotd
+
+    if test -e ~/.config/no-show-user-motd
+        mv ~/.config/no-show-user-motd ~/.config/uwelcome/disabled
+    end
```
hanthor, line 6: `~/.config/uwelcome/` isn't created anywhere in this PR, so if it doesn't already exist the `mv` fails. Combined with the missing `2>/dev/null` here (the profile.d version has it), that means: fish prints an `mv` error, the source file is never removed, and it repeats on **every** new fish shell — permanently, for exactly the users who had opted out.

```suggestion
    if test -e ~/.config/no-show-user-motd
        mkdir -p ~/.config/uwelcome
        mv ~/.config/no-show-user-motd ~/.config/uwelcome/disabled 2>/dev/null
    end
```

---

system_files/shared/etc/profile.d/uwelcome.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+if [ -e ~/.config/no-show-user-motd ]; then
+	mv ~/.config/no-show-user-motd ~/.config/uwelcome/disabled 2>/dev/null
+fi
```
hanthor, line 6: Same missing-directory problem as the fish version, but here `2>/dev/null` hides it — so instead of an error the opt-out just silently never migrates, and the `no-show-user-motd` file sticks around being re-tried on every login.

```suggestion
if [ -e ~/.config/no-show-user-motd ]; then
	mkdir -p ~/.config/uwelcome
	mv ~/.config/no-show-user-motd ~/.config/uwelcome/disabled 2>/dev/null
fi
```

(Kept the tab indentation to match the rest of the file.)

Worth double-checking against uwelcome upstream that `~/.config/uwelcome/disabled` is the path it actually reads — the entire migration hinges on it, and neither the code nor the tests currently pin it.

---

tests/test_motd_integration.bats
```diff
@@ -0,0 +1,74 @@
...
+    [ "${status}" -eq 0 ]
+    [ -f "${WORKDIR}/uwelcome.log" ]
+}
+
+
+# @test "uwelcome.sh: does not contain no-show-user-motd check" {
+#     run grep 'no-show-user-motd' "${UWELCOME_PROFILE}"
+#     [ "${status}" -ne 0 ]
+# }
+
```
hanthor: This one can't be restored as written — the migration shim legitimately references `no-show-user-motd` now. But rather than leaving it commented out, it'd be better to assert the shape you actually want, which also covers the `mkdir -p` fix above:

```suggestion
@test "uwelcome.sh: migrates legacy opt-out and still runs uwelcome" {
    touch "${HOME}/.config/no-show-user-motd"
    run bash "${UWELCOME_PROFILE}"
    [ "${status}" -eq 0 ]
    # Legacy marker is consumed, not left behind to retry every login
    [ ! -e "${HOME}/.config/no-show-user-motd" ]
    [ -e "${HOME}/.config/uwelcome/disabled" ]
    # uwelcome is invoked unconditionally; it owns the opt-out decision
    [ -f "${WORKDIR}/uwelcome.log" ]
}
```

This fails on the current code (the `mv` can't create the missing directory) and passes with the fix, so it pins the actual bug rather than just the absence of a string.

The equivalent for `fish_greeting.fish` is harder since fish isn't in CI — a grep-based check that `mkdir -p` appears inside the `if` block would at least catch a regression there.

While you're in this file: the header comment still says *"Neither profile.d nor fish_greeting should check for that file or gate the uwelcome invocation"*, which is no longer true of the first half. Worth rewording to describe the migration shim as intended behavior.

---

hanthor (commented):
Following up my earlier review with actual patches — the two `mkdir -p` suggestions are click-to-apply, and the third replaces one of the commented-out tests with one that pins the bug.

I verified the failure against the branch rather than reading it off the diff: `~/.config/uwelcome/` is not created by anything in this PR, so on a machine where uwelcome has never written its config dir, the `mv` fails in both shells. The fish path shows the error every shell start; the bash path swallows it and silently leaves the user opted *in* despite their earlier opt-out.

---

theMimolet:
Thanks for the fixes and tests, I didn't think about that
I restored the spaces by hand (i'm rather curious on which formatter you use)

As for the supply chain issue you mentioned a while back now, I've invited Ahmed and Jorge on the repos and so they'll be able to transfer them whenever.

ahmedadan:
> Thanks for the fixes and tests, I didn't think about that I restored the spaces by hand (i'm rather curious on which formatter you use)
> 
> As for the supply chain issue you mentioned a while back now, I've invited Ahmed and Jorge on the repos and so they'll be able to transfer them whenever.

I can help, can you go to settings and in the danger zone tell me if you can transfer ownership to the org? I don't have repository permissions to allow me to move them via GitHub.

theMimolet:
I can't transfer them either, I don't have any permissions in projectbluefin
image

ahmedadan:
> I can't transfer them either, I don't have any permissions in projectbluefin 

You'll need to give me more broad repository access then? I don't see the danger zone at all.

theMimolet:
I am probably missing something but I don't see how I could (?)
Otherwise I might need to transfer the repos to you first
image

ahmedadan:
Can you try inviting me as an admin?

ahmedadan:
I moved them both, thanks again for all the work here @theMimolet

theMimolet:
Thanks a lot ! 
I observed there's no role management for personal repos, so this was the best solution

hanthor (dismissed):
Re-review after the latest commits. The supply-chain blocker is resolved — repos are now under projectbluefin org. Approving with a few minor observations.

## Resolved since last review
- Supply chain: repos moved to projectbluefin/umotd and projectbluefin/uwelcome, pinned to specific commits
- Migration: legacy no-show-user-motd is now migrated (not deleted) to ~/.config/uwelcome/disabled
- Containerfile formatting: spaces restored

## Minor observations (non-blocking)

1. Test file header is slightly misleading. It says 'Neither profile.d nor fish_greeting should check for that file or gate the uwelcome invocation' — but both now DO check for it (to perform the one-time migration). The check is correct; the comment should say they should not gate invocation on it, or note that the check is transitional.

2. The old tips-presets included 'ublue' and 'ublue-dev' in addition to 'bluefin' and 'gnome'. The new tags.json only has ['bluefin', 'gnome']. Is dropping ublue/ublue-dev presets intentional? If uwelcome uses tags.json the same way the old config used tips-presets, this would be a behavior change.

3. The commented-out test 'fish_greeting: does not check no-show-user-motd' was valid for the old contract but the new contract does check (for migration). Consider replacing it with a test that verifies the fish migration behavior instead of leaving it commented.

4. color: 'auto' replaces use-accent-color: true — I assume this maps to the same behavior in uwelcome, just flagging for awareness.

None of these block merge. The core change is clean, the migration path preserves user opt-out state, and the build pins to org-owned repos.

theMimolet:
Regarding the observations : 
- the strings that were in the `ublue` tag now always show up no matter what and `ublue-dev` has been divided into the `vscode` and `containers` tags (so they can be added/removed with the devmode command) 
- the new `color` setting does replace `use-accent-color`, it allows pre-made themes the user can set manually (if auto doesn't work on their desktop and still want to customise it)

theMimolet:
Also for the test script - should we skip the fish test or should we add the fish package as deps for the tests ?  

Edit : Ahmed told me we can skip it.

castrojo:
Picked this up to land it as one complete feature. I pushed a merge of `main` plus fixes for the outstanding blockers — all CI is green and the branch is now `MERGEABLE`/`CLEAN`.

### Blockers fixed

**1. Build was pinned to mutable tags.** `Containerfile` had drifted back to `checkout tags/v0.3.1` / `tags/v0.3.4`. `docs/skills/containerfile.md` requires commit SHAs precisely because a tag can be repointed and silently change what we compile into every variant. Replaced with the SHAs those tags currently resolve to:

- umotd `c9df8ec6b53e9b2a644a6dc511fd6fde1baad08b`
- uwelcome `5280521bf21e14802d5a8bb1cffb942fa0b5efb7`

Both builder stages compiled clean from the SHAs, so the pins are good.

**2. `motd_toggle` is not a real translation key.** In `etc/uwelcome/config.json` it would have rendered the literal string `motd_toggle` in the banner instead of "Toggle this banner on/off". uwelcome v0.3.4's `docs/configuration.md` and its own `examples/bluefin.json` both use `banner_toggle`. Fixed.

**3. `ujust toggle-user-motd` vanished.** Restored as a one-line shim calling `uwelcome toggle`, per @hanthor's earlier note — existing muscle memory now works instead of erroring with "recipe not found".

### Test suite

The header comment in `test_motd_integration.bats` still described the *old* contract ("neither should check for that file"), which the migration shim deliberately violates, and the two assertions that contradicted it were commented out rather than rewritten. Replaced with live assertions of the actual contract — 8 new tests covering that the migration creates its target dir before the `mv`, consumes the legacy marker, is idempotent across logins, and never gates the `uwelcome` call. Added schema pinning for `config.json` and `tags.json` so an unknown translation key or tag fails CI instead of shipping.

Mutation-checked rather than assumed: reverting `banner_toggle` and deleting the `mkdir -p` each make the relevant tests fail. 20/20 pass.

### Verified upstream, not from memory

- `~/.config/uwelcome/disabled` is genuinely the path uwelcome checks (`internal/state/state.go`), so the migration preserves opt-out.
- @hanthor's earlier question about dropping `ublue`/`ublue-dev` from the tag list: those aren't valid tags in umotd v0.3.1 anymore. The supported set is `aurora`, `bazzite`, `bazzite-deck`, `bluefin`, `gnome`, `kde`, `vscode`, `containers` — so `["bluefin", "gnome"]` is correct, and the new test pins it.
- `color: "auto"` is a documented uwelcome option that maps to the system accent color, matching the old `use-accent-color: true`.

### Docs

Updated `docs/skills/containerfile.md` for the two builder stages and the umotd/uwelcome config split, and corrected the stale `umotd.sh` exemption row in `docs/TESTING.md`.

Pre-existing failures in `test_changelog.bats` (11) and `test_update_just.bats` (4) are unrelated — identical counts on `main`.

Thanks @theMimolet, nice work on the split. 🎉

castrojo (approved):
Approving. All prior blockers are resolved and verified against upstream source rather than assumed: the build is pinned to immutable commit SHAs (both stages compiled clean), the banner uses the real `banner_toggle` translation key, the opt-out migration targets the path uwelcome actually reads, and `ujust toggle-user-motd` still works via a shim.

Test coverage is now stronger than what it replaced — the commented-out assertions are live again, rewritten against the migration contract as it really is, plus schema pinning so a bad translation key or tag fails CI instead of shipping. 20/20 pass and the key assertions were mutation-checked.

Full CI green. Landing it.

PR #852, opened 2026-07-23
feat(brew): manage cask lifecycle in brew-preinstall

castrojo:
## What

Extends the brew-preinstall managed lifecycle to `cask "..."` lines in `preinstall.d/*.Brewfile`:

- casks are installed by the existing `brew bundle` pass and tracked in the state file under a new `casks` key
- dropping a cask line uninstalls it on next login via `brew uninstall --cask` (same only-if-installed, never-touch-user-installs safety rules as formulae)
- legacy state files without a `casks` key read as an empty list — no migration
- fixes a latent pipefail bug: `grep` exits 1 when the Brewfile set declares no `brew` (or `cask`) lines, aborting the script under `set -e`
- filters empty strings out of the persisted package/cask arrays

## Why

Prepares for shipping ChairLift as a preinstalled cask (frostyard/homebrew-tap#1, frostyard/chairlift#53). Follow-up PR adds the Brewfile + config.

## Validation

- `bats tests/test_brew_preinstall.bats` — 21/21 pass (6 new cask tests)
- `just check`, `pre-commit run` on changed files — pass
- brew-lifecycle skill updated in the same change (AGENTS.md rule)

Blast radius: `system_files/shared/` — behavior for existing Brewfiles is unchanged (verified by the pre-existing 15 tests).

## Summary by CodeRabbit

* **New Features**
  * Brew lifecycle management now tracks both formulae and cask applications.
  * Removed casks are automatically uninstalled during reconciliation.
  * State records now include managed casks, with compatibility for older state files.
  * Added guidance for adding and verifying casks.

* **Bug Fixes**
  * Improved handling when Brewfiles contain no casks.
  * Prevented similarly named formulae and casks from affecting each other.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install test deps
         run: |
           sudo apt-get install -y bats
-          pip install pytest pytest-cov
+          pip install pytest pytest-cov pyyaml
 
       - name: Run shellcheck — extensionless scripts
         run: |
@@ -53,6 +53,9 @@ jobs:
 
       - name: Run pytest (hooks.py)
         run: python3 -m pytest tests/test_hooks.py -v --cov=system_files/bluefin/etc/bazaar --cov-report=term-missing --cov-report=xml:coverage.xml --cov-report=html:coverage-html --cov-fail-under=80
+
+      - name: Run pytest (ChairLift maintainer config)
+        run: python3 -m pytest tests/test_chairlift_config.py -v
 
       - name: Upload coverage report
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/.github/workflows/validate-chairlift-config.yaml
+++ b/.github/workflows/validate-chairlift-config.yaml
@@ -1,0 +1,43 @@
+name: Validate ChairLift Config
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    paths:
+      - "system_files/shared/usr/share/chairlift/**"
+      - "tests/check-chairlift-config"
+      - ".github/workflows/validate-chairlift-config.yaml"
+  push:
+    branches:
+      - main
+    paths:
+      - "system_files/shared/usr/share/chairlift/**"
+      - "tests/check-chairlift-config"
+  # Upstream can change its schema without us touching anything, and an
+  # unknown key disables the whole app. Catch that drift on a schedule
+  # instead of on the next unrelated PR.
+  schedule:
+    - cron: "17 6 * * 1"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && format('pr-{0}', github.event.pull_request.number) || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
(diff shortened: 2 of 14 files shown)


hanthor (dismissed):
Approving. The cask lifecycle mirrors the existing formula handling cleanly: install via `brew bundle`, track under a new `casks` state key, and `brew uninstall --cask` only when `brew list --cask` confirms it — same never-touch-user-installs safety as formulae. Legacy state without a `casks` key degrades to an empty list via `.casks[]? // empty`, and the empty-string filtering with `map(select(length > 0))` is correct.

The switch from `grep '^brew "'` to `sed -n '...p'` is a genuine fix: `grep` exits 1 on no-match, which under `set -e` would abort (or with `|| true` mask real errors and risk an empty current set being read as "uninstall everything"). Gating the removal/state-write phase behind `bundle_failed` so a single failing Brewfile forces a full retry on next login is the right call for a content-addressed service.

21/21 bats pass, the 6 new tests cover the meaningful edges (legacy state, same-name formula/cask, bundle-failure skips removals). Unit Tests, Validate, and both arch builds are green; only the long-running E2E image compose is still in progress. LGTM.

castrojo:
Independent code-review pass completed (dedicated review agent + adversarial testing of the removal path). Findings and resolutions, all now in the branch:

1. (major, fixed) Failed uninstalls were absorbed by a warning while fresh state was still written — the hash fast-exit would then skip the removal forever. Removal failures now exit 1 before the state write, so the old state survives and systemd/next login retries. Covered by a new bats test.
2. (minor, fixed) A corrupt state file was silently treated as an empty previous set. Behavior stays fail-safe (no removals on bad data, state self-heals on the next write) but it now logs a warning. Covered by a new bats test.
3. (hardening, earlier commit) Managed-name extraction moved from grep|sed||true to sed -n: no-match exits 0 but real I/O errors still fail, so an unreadable Brewfile can never produce an empty current set that the diff would treat as uninstall-everything.
4. (hardening, earlier commit) Per-Brewfile bundle failure isolation: one unreachable external tap no longer blocks the rest; any failure exits 1 before removals/state-write.

25/25 bats, shellcheck clean. Adversarial checks run by hand: corrupt state JSON, null casks value, cask-only Brewfile set, mock uninstall failures.

hanthor (approved):

hanthor (changes requested):
The image build check ('Build and push image (x86_64)' & 'aarch64') failed. Please rebase or fix the image build failure before this PR can be merged.

castrojo:
Rebased onto current `main` (was `CONFLICTING`). The only conflicts were in the generated `docs/skills/index.json`/`index.md`, resolved by regenerating with `scripts/generate_skill_index.py --write` rather than hand-merging.

No behaviour change to `brew-preinstall` in this push — the five commits are the same work, replayed.

Local verification on the rebased head:
- `bats tests/test_brew_preinstall.bats` — 25/25 pass
- `shellcheck system_files/shared/usr/libexec/brew-preinstall` — clean
- `just check` — clean
- `pre-commit run --all-files` — all hooks pass
- `scripts/generate_skill_index.py --check` — index up to date

`mergeable` is now `MERGEABLE`. @hanthor this should clear the build failure that prompted `CHANGES_REQUESTED` — re-review when you get a chance.

castrojo:
Superseded by #981, which preserves Edward's contributor-owned PR and now contains the hardened cask lifecycle, expanded tests, and current brew lifecycle documentation. Thank you to everyone who developed and reviewed the earlier implementation here. The separate ChairLift rollout remains tracked in #854.

Issue #102, opened 2026-01-02
Refactor dx mode vscode extensions


castrojo:
https://github.com/ublue-os/bluefin/blob/1e2c836ee21e193baf4bb3bfe85b82eec217ffd2/system_files/dx/usr/share/ublue-os/user-setup.hooks.d/10-vscode.sh#L5

^^^ we should use the `vscode` things in the Brewfiles to do this, then all we need to ship for vscode is the settings.json and that would be much simpler!

castrojo:
https://podman-desktop.io/blog/2025/05/05/vs-code-with-podman-desktop

^^^ We should ship this out of the box while we're add it, we ship all of these already, adding that podman extension would be baller!

PR #853, opened 2026-07-23
feat(brew): preinstall ChairLift with Bluefin maintainer config

castrojo:
## What

Ships [frostyard/chairlift](https://github.com/frostyard/chairlift) on all Bluefin variants via the managed preinstall set:

- `preinstall.d/chairlift.Brewfile` — `tap "frostyard/tap", trusted: true` + `cask "chairlift"`
- `/usr/share/chairlift/config.yml` maintainer defaults for Bluefin:
  - bootc staged updates enabled through the fixed image-side helper and authenticated PolicyKit action
  - updex (`features_page.features_group`) remains disabled because Bluefin does not ship an updex helper
  - `bundles_paths` pointed at `/usr/share/ublue-os/homebrew`
  - help links pointed at Bluefin docs/issues/ask resources
  - update scheduling knobs disabled so Bluefin keeps its existing `uupd` policy
- `/usr/libexec/bootc-update-stage` stages only with `bootc upgrade --quiet --download-only`; it never applies or reboots
- the PolicyKit action pins the helper path and requires `auth_admin`; no passwordless rule is added
- `tests/test_chairlift_config.py` pins those decisions

## Why this matters

This is the first shared managed-preinstall cask rollout for a GUI app on Bluefin. It codifies Bluefin policy so ChairLift does not compete with the existing update model, while the resolved polkit integration provides authenticated staging without adding a new scheduling or reboot path.

## Review focus

Please review the policy decisions and the shared config shape:

- the fixed helper and authenticated PolicyKit action
- bootc staging enabled while updex remains disabled
- the hidden update scheduling controls
- the bundle path/help links and regression tests

## Current status

The branch is rebased onto the current `feat/brew-preinstall-casks` base from #852, with the merged #892 follow-up included. Local validation passed for the 25-test brew-preinstall Bats suite, 9 ChairLift pytest tests, catalog freshness, `just check`, and `pre-commit run --all-files`. The aggregate `just test` run reaches the known network-dependent changelog Bats failures in this environment; no changed-test failure was observed.

GitHub CI will recompute against the rebased head. The PR remains open for the required maintainer security review of the privileged PolicyKit path; this update does not merge or self-approve the PR.

## Stacked on

#852 (cask lifecycle in brew-preinstall) — do not retarget to main before it merges.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install test deps
         run: |
           sudo apt-get install -y bats
-          pip install pytest pytest-cov
+          pip install pytest pytest-cov pyyaml
 
       - name: Run shellcheck — extensionless scripts
         run: |
@@ -53,6 +53,9 @@ jobs:
 
       - name: Run pytest (hooks.py)
         run: python3 -m pytest tests/test_hooks.py -v --cov=system_files/bluefin/etc/bazaar --cov-report=term-missing --cov-report=xml:coverage.xml --cov-report=html:coverage-html --cov-fail-under=80
+
+      - name: Run pytest (ChairLift maintainer config)
+        run: python3 -m pytest tests/test_chairlift_config.py -v
 
       - name: Upload coverage report
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/.github/workflows/validate-chairlift-config.yaml
+++ b/.github/workflows/validate-chairlift-config.yaml
@@ -1,0 +1,43 @@
+name: Validate ChairLift Config
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    paths:
+      - "system_files/shared/usr/share/chairlift/**"
+      - "scripts/check-chairlift-config.py"
+      - ".github/workflows/validate-chairlift-config.yaml"
+  push:
+    branches:
+      - main
+    paths:
+      - "system_files/shared/usr/share/chairlift/**"
+      - "scripts/check-chairlift-config.py"
+  # Upstream can change its schema without us touching anything, and an
+  # unknown key disables the whole app. Catch that drift on a schedule
+  # instead of on the next unrelated PR.
+  schedule:
+    - cron: "17 6 * * 1"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && format('pr-{0}', github.event.pull_request.number) || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Install deps
+        run: pip install pyyaml
+
+      - name: Validate ChairLift config against upstream schema
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: python3 scripts/check-chairlift-config.py
--- a/.github/workflows/validate-chairlift-config.yaml
+++ b/.github/workflows/validate-chairlift-config.yaml
@@ -7,14 +7,14 @@ on:
   pull_request:
     paths:
       - "system_files/shared/usr/share/chairlift/**"
-      - "scripts/check-chairlift-config.py"
+      - "tests/check-chairlift-config"
       - ".github/workflows/validate-chairlift-config.yaml"
   push:
     branches:
       - main
     paths:
       - "system_files/shared/usr/share/chairlift/**"
-      - "scripts/check-chairlift-config.py"
+      - "tests/check-chairlift-config"
   # Upstream can change its schema without us touching anything, and an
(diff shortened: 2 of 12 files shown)


castrojo:
Update-policy addition: `updates_settings_group` is now disabled too. Bluefin's update policy belongs to uupd (silent background staging via 6-hour timer + AC-connect trigger, no notifications; user reboots on their own schedule). ChairLift must not surface competing scheduling knobs. Verified from upstream source that ChairLift ships no autostart/daemon/desktop-notification paths — it cannot interrupt users while closed. Pinned by `test_update_policy_stays_with_uupd`.

castrojo:
Follow-on work is now tracked in issues: #854 (undraft/rollout sequence once the upstream release lands) and #855 (re-enable bootc staging/updex when frostyard/chairlift#54 resolves — security review gate).

castrojo:
Updated the branch with a small documentation note clarifying the ChairLift tap dependency for reviewers. Local regression tests pass. The remaining CI issue is the upstream Homebrew tap/cask publication requirement for frostyard/tap and chairlift, which still needs to land before the Brewfile validation can pass.

hanthor (approved):
Approving — this is careful work and the reasoning in the comments is the good kind.

Specifically worth calling out:

- **`grep` → `sed -n` for the current set.** The comment explains the actual hazard: `grep` exits 1 on no-match, `|| true` would mask real errors, and an empty `current_packages` feeding the `comm -23` diff means "uninstall everything we manage." `sed -n` exiting 0 on no-match is the correct primitive here.
- **Corrupt state file is fail-safe.** Empty previous sets → no removals, plus an explicit warning rather than silently swallowing it, and the state write self-heals the file.
- **Bailing before the state write on any bundle or uninstall failure.** This is the subtle one — writing the new hash after a failed removal would make the fast-exit path skip the retry forever. Getting that ordering right matters more than it looks.
- `map(select(length > 0))` on the jq arrays avoids the empty-string-element artifact from `echo ""  | jq -R`.

Test coverage matches the behavior claims one-for-one, including the legacy state file with no `casks` key. No changes requested.

---

castrojo:
## Ship-blocking bug found and fixed

A holistic review of this feature found that the config this PR ships **would have disabled ChairLift entirely for every user**.

### The bug

`config.yml` declared `updates_page.updates_settings_group` to express "uupd owns update scheduling". **That group has never existed upstream.** `internal/config/config.go::defaultConfig()` defines `updates_page` as exactly four groups: `bootc_updates_group`, `flatpak_updates_group`, `brew_updates_group`, `brew_trust_group`.

ChairLift does not ignore unknown keys — it fails closed:

```
parseAndValidate  → LoadError{Kind: KindSchema}   (validate.go: unknown group ⇒ KindSchema)
Load()            → disabledConfig()              (config.go)
disabledConfig()  → Enabled=false on EVERY group, EVERY page
LoadError         → persistent "Configuration error … All feature groups are disabled" toast
```

Since we preinstall ChairLift silently at next login, this shipped an empty, broken app to every `bluefin`, `bluefin-lts`, and `dakota` user simultaneously. Nobody opted in.

Fixed by deleting the key and putting the uupd rationale in a comment — which is where it belonged, since there is no scheduling UI to suppress.

### Why CI did not catch it

Three independent gaps, all now closed:

1. **The test validated our YAML against itself.** `KNOWN_GROUPS` was a hand-written allowlist edited in this same PR to include the invented group, so `test_config_parses_and_uses_known_pages_and_groups` passed. `test_update_policy_stays_with_uupd` was vacuous for the same reason — it asserted our own file said what we wrote. Replaced with a strict-subset assertion.
2. **`tests/test_chairlift_config.py` was wired into neither `just test` nor `unit-tests.yml`** — the tests this PR added never ran anywhere. Both now run it.
3. **No check against upstream truth.** Added `scripts/check-chairlift-config.py` + `Validate ChairLift Config` workflow: fetches upstream `config.yml` (a complete enumeration kept in lockstep with `defaultConfig()`) and fails on any key we use that upstream does not define. Runs on config PRs **and weekly**, since upstream can drift without us touching anything.

Verified the guard actually works — reintroducing the bug fails the drift check and 3 pytest cases; removing it passes all 10.

### Also fixed

**Dropped `--quiet` from `/usr/libexec/bootc-update-stage`.** `bootc.StageUpdate` merges the helper stdout+stderr and streams each line into ChairLift progress view, so `--quiet` left the user watching an empty dialog for the length of an image pull.

### Reconfirmed as correct

- Upstream is fully unblocked: frostyard/chairlift#53 and #102 merged, #54 closed, frostyard/homebrew-tap#1 merged.
- Upstream explicitly expects distros to supply the stage script themselves — the `chairlift-system-integration` package ships the polkit policies but *not* the helper. This PR does exactly that.
- `bootc upgrade --download-only` is right for Bluefin. Upstream avoids plain `bootc upgrade` only because of a snosi-specific composefs registry-transport bug that does not apply here.

### Known caveat (deliberate)

The tap cask still targets the rolling `dev` tag with `sha256 :no_check`, so the preinstalled payload is mutable and unverified even though v0.10.1 exists with per-arch checksums. Accepted for the initial rollout; tracked as a follow-up.

### Still required before merge

- Human **security review** of the privileged `pkexec` → `/usr/libexec/bootc-update-stage` path and the pinned polkit action.
- Real-hardware validation, specifically **confirming no configuration-error toast**.
- Lab e2e on `bluefin:testing`.

Rebased onto the rebased #852. Do not retarget to `main` before #852 merges.

ahmedadan (approved):

PR #862, opened 2026-07-25
fix(ujust): install fzf before choosing recipes

### LINKED ISSUE #861: `ujust --choose` crashes on Bluefin because `fzf` isn't installed
### Before submitting

### ujust report gist URL

https://gist.github.com/znmeb/079abe49e4b5374f7a1f554f17367024

### What happened?

❯ ujust --choose
sh: line 1: fzf: command not found
error: chooser `fzf --multi --preview 'just --unstable --color always --justfile "/usr/share/ublue-os/just/00-entry.just" --show {}'` failed: exit status: 127

System is on the new Bluefin NVIDIA image:

`ghcr.io/projectbluefin/bluefin-nvidia:testing`

### Steps to reproduce

Run `ujust --choose`

### Affected area

brew / CLI tools

### Extra context

_No response_

castrojo:
## Summary
- install the Homebrew-managed `fzf` on demand when `ujust --choose` runs before brew-preinstall
- preserve existing behavior for all other ujust commands
- add Bats coverage for the on-demand install path

Fixes #861

## Validation
- `bats tests/test_ujust.bats`
- `just check`
- `just test`
- `shellcheck -S warning system_files/shared/usr/bin/ujust`
- `actionlint` workflow validation

`pre-commit run --all-files` is currently blocked by pre-existing untracked docs artifacts (`docs/skills/INDEX.md` front matter/index-link checks).

## Summary by CodeRabbit

* **New Features**
  * Automatically installs the `fzf` utility when using `ujust --choose` and it isn’t available.
  * Continues supporting standard `ujust` commands without installing unnecessary tools.

* **Tests**
  * Added coverage verifying on-demand `fzf` installation and correct command execution.

### DIFF
--- a/system_files/shared/usr/bin/ujust
+++ b/system_files/shared/usr/bin/ujust
@@ -1,3 +1,20 @@
 #!/usr/bin/bash
 
-just --justfile /usr/share/ublue-os/just/00-entry.just "${@}"
+if [[ " ${*} " == *" --choose "* ]] && ! command -v fzf >/dev/null 2>&1; then
+    BREW_BIN="$(command -v brew 2>/dev/null || true)"
+    BREW_BIN="${BREW_BIN:-/var/home/linuxbrew/.linuxbrew/bin/brew}"
+
+    if [[ ! -x "${BREW_BIN}" ]]; then
+        echo "ujust: fzf is missing and Homebrew was not found" >&2
+        exit 127
+    fi
+
+    echo "ujust: installing missing fzf with Homebrew..." >&2
+    if ! "${BREW_BIN}" install fzf; then
+        echo "ujust: failed to install fzf" >&2
+        exit 1
+    fi
+    eval "$("${BREW_BIN}" shellenv)"
+fi
+
+exec just --justfile /usr/share/ublue-os/just/00-entry.just "${@}"
--- a/system_files/shared/usr/bin/ujust
+++ b/system_files/shared/usr/bin/ujust
@@ -1,7 +1,15 @@
 #!/usr/bin/bash
 
-if [[ " ${*} " == *" --choose "* ]] && ! command -v fzf >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


hanthor (changes requested):
The on-demand path needs to fail closed before dispatching just. The current joined-argument search treats text containing --choose as the option, and a failed or ineffective brew shellenv can still fall through to just with fzf unavailable. Please match --choose as an exact argument, propagate shellenv failure, verify fzf is actually resolvable after applying the environment, and cover those failure paths with an isolated PATH in Bats.

hanthor (dismissed):
Approving. The ordering is correct and well-tested — install, then `shellenv`, then `exec just`, with each failure mode exiting distinctly (127 for missing brew / still-missing fzf, 1 for install or shellenv failure) and never falling through to `just`. Switching to `exec` is a nice incidental cleanup. The bats suite covers all five paths including the negative ones.

One behavioral note, not a blocker: `ujust --choose` now silently installs software into the user's Homebrew prefix as a side effect of a command that reads like "show me a menu." The `installing missing fzf` message goes to stderr, so it is at least visible. If you'd rather not have an implicit install, a `read -p` confirmation would be the alternative — but given `--choose` is unusable without fzf, auto-installing is defensible. Your call.

---

castrojo:
Superseded by #902, which is now rebased and queued.

The `ujust` change here and in #902 are byte-identical (verified: same SHA256 over the diff hunk for `system_files/shared/usr/bin/ujust`). #902 is a strict superset — it carries the same fix plus:

- `docs/skills/brew-lifecycle.md` documentation for the bootstrap behavior
- Justfile wiring so the chooser-bootstrap test runs in the standard suite

Closing this one to keep a single source of truth for the fix. No work is lost — the code landing in #902 is exactly what was written here.

Also resolves the duplicate-issue situation: #861 and #899 describe the same bug; #902 closes #899.

PR #892, opened 2026-07-29
feat(brew): re-enable ChairLift bootc staging via system-integration split

### LINKED ISSUE #855: brew: re-enable ChairLift bootc staging and updex when upstream polkit integration resolves
What: The ChairLift maintainer config shipped in PR #853 (system_files/shared/usr/share/chairlift/config.yml) disables bootc_updates_group and features_group (updex). Reason: the upstream polkit policy hardcodes Snow's /usr/libexec/bootc-update-stage, and polkit only reads /usr/share/polkit-1 — unreachable from a user-scope Homebrew cask. updates_settings_group is also disabled so no in-app scheduling knobs compete with uupd's silent background update policy (that one stays off regardless).

Fix: When frostyard/chairlift#54 resolves (options proposed upstream: configurable stage-script path, system-integration split package, or upstreamed stage script):
1. Ship whatever image-side glue the chosen direction requires (polkit policy/rules under /usr/share/polkit-1, stage script under /usr/libexec). Security-sensitive: privileged pkexec path — requires human review before merge.
2. Flip bootc_updates_group (and features_group only if updex ever ships on Bluefin) in config.yml.
3. Update tests/test_chairlift_config.py: test_polkit_dependent_groups_stay_disabled pins the current off state and will fail loudly when flipped — update it to pin the new decision.
4. Confirm the staged-update flow does not prompt users: silent staging, reboot applies on the user's schedule (uupd remains the policy owner).

Automatable: no (security review gate on the polkit/pkexec glue)

castrojo:
Closes #855.

## What
frostyard/chairlift#54 (polkit integration blocker) resolved via frostyard/chairlift#102, which shipped **option (b)** from the issue: a `frostyard-chairlift-system-integration` package with fixed privileged paths, rather than a configurable stage-script path or fully upstreamed script. The distro (Bluefin) is still responsible for providing `/usr/libexec/bootc-update-stage` itself — upstream does not ship one.

This PR ships that image-side glue and flips the config flag per issue #855's step 2/3:

- **`system_files/shared/usr/share/polkit-1/actions/org.frostyard.ChairLift.bootc.policy`** — verbatim copy of upstream's bootc PolicyKit action (`auth_admin` / `auth_admin_keep`, no passwordless rule — matches ChairLift's own "no passwordless PolicyKit rules" posture).
- **`system_files/shared/usr/libexec/bootc-update-stage`** — new privileged helper, invoked only via the polkit action above (pkexec pins the exec path). It's a thin wrapper: `bootc upgrade --download-only`. It intentionally never applies/reboots — staging only. Applying still belongs to the user's own reboot, or uupd's existing background policy; this does not add any new update-scheduling behavior.
- **`config.yml`** — `bootc_updates_group.enabled: true`. `features_page.features_group` (updex) stays **disabled**: no updex helper ships on Bluefin yet, independent of this fix, so that part of #855 is intentionally left for a future PR.
- **`tests/test_chairlift_config.py`** — replaced `test_polkit_dependent_groups_stay_disabled` (now pins the opposite decision) with tests asserting: bootc staging is enabled, updex stays disabled, the polkit policy pins the fixed helper path with `auth_admin`, and the stage script is executable and only ever invokes `bootc upgrade --download-only` (no `--apply`).
- **`docs/skills/brew-lifecycle.md`** — updated to describe the resolved state instead of "disabled until #54 resolves."

## Why this is safe
- No passwordless PolicyKit rule is added — every staging invocation still requires `auth_admin` (the user authenticates), matching upstream's default and README's "ChairLift does not install passwordless PolicyKit rules" statement.
- The helper script takes no arguments and only ever execs one fixed, non-interactive `bootc upgrade --download-only` command — it cannot be used to run arbitrary bootc subcommands even if pkexec forwards args.
- Staging never applies/reboots — uupd remains the sole owner of when an update actually takes effect, matching Bluefin's "silent staging, user reboots on their own schedule" policy from issue #853/#855.

## Base branch note
This targets `feat/chairlift-preinstall` (PR #853), not `main` — `config.yml` and `tests/test_chairlift_config.py` only exist on that branch today (PR #853 is still open). This PR should land after/alongside #853, whichever order the maintainers prefer.

## Security review
Per `docs/skills/human-gates.md`, this touches a privileged pkexec/PolicyKit path and needs explicit maintainer security review before merge — flagging that here rather than self-approving.

## Testing
- `python3 -m pytest tests/test_chairlift_config.py -v` — 9/9 pass (full `tests/` pytest suite: 79/79 pass).
- Validated the PolicyKit XML parses (`xml.dom.minidom`).
- `shellcheck`/`bats` were not available in this sandbox; the new script has no unmanaged inputs (no args processed) and was reviewed by hand against the existing `brew-preinstall` libexec script's style.

Assisted-by: Claude Sonnet 5 via GitHub Copilot

### DIFF
--- a/docs/skills/brew-lifecycle.md
+++ b/docs/skills/brew-lifecycle.md
@@ -191,13 +191,19 @@ cask "chairlift"
 ChairLift is a GTK4 system-management GUI (Homebrew, Flatpak, bootc).
 Its Bluefin maintainer config ships at
 `system_files/shared/usr/share/chairlift/config.yml` (admin override:
-`/etc/chairlift/config.yml`); bootc staging and updex stay disabled
-until frostyard/chairlift#54 resolves the polkit integration. Guarded
-by `tests/test_chairlift_config.py`. The cask is now reviewable from the
-common side because the upstream ChairLift release assets include the needed
-`data/` files and arm64 tarballs; the remaining dependency for full Brewfile
-validation is the upstream Homebrew tap PR publishing the cask to the
-Homebrew resolvers that CI uses.
+`/etc/chairlift/config.yml`). frostyard/chairlift#54 resolved via the
+system-integration split (frostyard/chairlift#102): Bluefin now ships the
+fixed `/usr/libexec/bootc-update-stage` helper and the bootc PolicyKit
+policy under `system_files/shared/usr/share/polkit-1/actions/`, so
+`bootc_updates_group` is enabled (staging only —
+`bootc upgrade --download-only`; applying stays with the user's own
+reboot or uupd's background policy). `features_group` (updex) stays
+disabled — no updex helper ships on Bluefin yet, independent of the
+polkit fix. Guarded by `tests/test_chairlift_config.py`. The cask is now
+reviewable from the common side because the upstream ChairLift release
+assets include the needed `data/` files and arm64 tarballs; the remaining
+dependency for full Brewfile validation is the upstream Homebrew tap PR
+publishing the cask to the Homebrew resolvers that CI uses.
 
 ### Add a variant-specific Brewfile
 
--- a/system_files/shared/usr/libexec/bootc-update-stage
+++ b/system_files/shared/usr/libexec/bootc-update-stage
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# bootc-update-stage — privileged helper invoked via ChairLift's fixed
+# org.frostyard.ChairLift.bootc.stage polkit action (pkexec exec.path is
+# pinned to this exact path; see
+# system_files/shared/usr/share/polkit-1/actions/org.frostyard.ChairLift.bootc.policy).
+#
+# Downloads and stages the next bootc system image update. It intentionally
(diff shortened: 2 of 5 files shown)


hanthor (approved):
Approving. The privileged-helper shape is right:

- `exec.path` in the polkit action is pinned to the exact helper path, and the helper genuinely ignores `$@` — it `exec`s a fixed `bootc upgrade --quiet --download-only` with no argument passthrough, so this can't be turned into "run arbitrary bootc subcommands as root."
- `auth_admin` / `auth_admin_keep` rather than `yes`, and the test asserts there's no passwordless bypass.
- Staging only, no `--apply`, so uupd stays the owner of when an update actually takes effect. The test asserting exactly one `exec` line with `--download-only` and no `--apply` is a good guard against that drifting.

Keeping `features_group` (updex) disabled for a *different* reason than the polkit fix, and saying so in both the config comment and the test docstring, is the right call — those two were previously conflated.

One logistical note: this stacks on #853 (it edits `tests/test_chairlift_config.py` and the config that PR introduces), so merge order matters.

---

PR #908, opened 2026-08-01
chore(docs): generate the parity matrix and fix phantom workflow docs

castrojo:
## What

- Adds `scripts/generate_parity_matrix.py` — generates the factory parity matrix in `docs/factory/README.md` from live GitHub state.
- Removes three workflows from `docs/skills/workflow-map.md` that are documented as existing in `common` but exist nowhere.
- Documents `unit-tests.yml`, which existed but was undocumented.

## Why

Retiring `skill-drift.yml` (#previous) exposed that the hand-maintained parity matrix had drifted into fiction — it marked `skill-drift.yml` present for `common`, which never had it.

Auditing `workflow-map.md` against `ls .github/workflows/` found the same rot:

| Documented in `common` | Actually exists? |
|---|---|
| `lifecycle-caller.yml` | **No** — and neither does the `actions/lifecycle.yml` it claims to call. Lifecycle lives in `bonedigger`, consumed via `bonedigger.yml`. |
| `sync-codeowners.yml` | **No** — exists in no factory repo. |
| `backfill-pipeline.yml` | **No** — and it was listed *both* as live and as "must not be re-added". |
| `unit-tests.yml` | Yes, but was undocumented. |

## Regenerating also corrected errors nobody had noticed

The hand-written table also had these wrong:

- Renovate marked ❌ for `dakota` and ❓ for `bluefin-lts` — both have it at `.github/renovate.json5`.
- pre-commit marked — for `actions` — it is present.

## Verification

```bash
# workflow-map now matches reality in both directions
comm -13 <(ls .github/workflows/ | sort) <(grep -oE '^\| `[a-z0-9.-]+\.ya?ml`' docs/skills/workflow-map.md | tr -d '|` ' | sort)
comm -23 <(ls .github/workflows/ | sort) <(grep -oE '^\| `[a-z0-9.-]+\.ya?ml`' docs/skills/workflow-map.md | tr -d '|` ' | sort)
# both empty

GITHUB_TOKEN=$(gh auth token) python3 scripts/generate_parity_matrix.py --check
```

`just check` and `pre-commit run --all-files` pass.

## Note

The script is **not** wired into pre-commit — it makes network calls, which would break offline commits and burn rate limit. Per `agentic-model.md` it is also not a CI gate. Run manually or on a schedule.


castrojo:
Closing — the script was the wrong fix.

`docs/factory/README.md` already says, two sections below the matrix:

> Factory gaps are tracked as GitHub issues — **not in this doc**. Query GitHub for the live state.

The parity matrix was live state embedded in a doc, violating the file's own convention on the same page. A hand-maintained table drifts; a generated one is a table nobody remembers to regenerate — I had even written "run it manually or on a schedule," which is the same failure mode.

So the matrix is deleted rather than generated, replaced with the query pattern the file already uses. The genuinely useful part of this PR — the phantom workflow fixes in `workflow-map.md` — landed as a doc-only change in 057d213.

PR #929, opened 2026-08-04
ci: reclaim runner disk before image builds

### LINKED ISSUE #796: ci: ISO builds fail intermittently due to ENOSPC during OSTree extraction
What: The ISO build workflow (build-iso-bluefin.yml) intermittently fails with ENOSPC on GitHub-hosted runners when building non-composefs (OSTree) variants like `bluefin-lts-hwe`. These variants unpack all ~120 layers of their payload image onto the runner, which can exhaust the default disk space.\n\nFix: In projectbluefin/dakota-iso, we switched to aggressively freeing disk space at the start of the job using the org's standard `ublue-os/remove-unwanted-software@v9` action, which reclaims ~30 GB. We should evaluate if this needs to be applied to other workflows producing OSTree artifacts.\n\nAutomatable: yes

castrojo:
## Summary

- Reclaim hosted-runner disk space immediately after checkout in the common image build workflow.
- Use the org-standard `ublue-os/remove-unwanted-software@v10` action with `extra-squeeze: "true"`.
- Document the disk-space preparation pattern and source-verification commands in `docs/skills/ci.md`.

Closes #796

## Evidence

Issue #796 identifies intermittent `ENOSPC` while large OSTree/OCI payloads are extracted on GitHub-hosted runners. The shared `projectbluefin/actions/bootc-build/setup-runner` action already uses this exact cleanup action and pin. The new step applies the same source-backed mitigation before this repository's image build.

## Validation

- `git diff --check` — passed.
- Workflow assertions — passed: cleanup step is after checkout and before tag generation, uses the verified v10 pin, and sets `extra-squeeze: "true"`.
- `just check` — not run; `just` is unavailable in this environment.
- `pre-commit run --all-files` — not run; `pre-commit` is unavailable in this environment.
- Full image build — not run; this is a workflow-only change.


bketelsen:
I audited the actual ISO workflows for #796. This workflow already calls `projectbluefin/actions/bootc-build/setup-runner`, whose current implementation runs `remove-unwanted-software` v10 with `extra-squeeze`, so adding the action here duplicates cleanup in the regular Common image build and does not cover ISO extraction.

The remaining uncovered path is `projectbluefin/iso`’s reusable Titanoboa workflow; I opened projectbluefin/iso#69 with the cleanup immediately before the ISO job’s dependency installation and checkout.

Separately, this PR’s validation failure is from `docs/skills/ci.md`: it has no required front matter and is missing from `docs/SKILL.md` and `docs/skills/index.json`.

castrojo:
Closing as already covered upstream — the disk reclaim this adds is a duplicate of one that already runs.

The `Setup runner` step immediately above the insertion point resolves to `projectbluefin/actions/bootc-build/setup-runner`, which already runs the identical action, at the identical SHA, with the identical input:

```yaml
# projectbluefin/actions bootc-build/setup-runner/action.yml:41
- name: Maximize build space (remove-software)
  uses: ublue-os/remove-unwanted-software@695eb75bc387dbcd9685a8e72d23439d8686cba6 # v10
  with:
    extra-squeeze: "true"
```

Re-derive with:

```bash
gh api repos/projectbluefin/actions/contents/bootc-build/setup-runner/action.yml \
  --jq .content | base64 -d | grep -n -A5 -B3 'Maximize build space'
```

So the space is already reclaimed before the added step runs; the second invocation would find little to remove while still costing runner time on every build, across both arches.

Two smaller notes for anyone revisiting this:

- The added `docs/skills/ci.md` says *"Keep this step immediately after checkout"*, but the diff places it after `Setup runner`. If this ever is re-added, doc and diff should agree.
- #796 was specifically about **ISO** builds (`build-iso-bluefin.yml`, non-composefs OSTree variants unpacking ~120 layers). `common` has no ISO workflows, so `build.yml` was not the workflow experiencing ENOSPC.

The SHA pin itself was correct and correctly commented — the instinct was right, it just already happens via the shared action.

PR #932, opened 2026-08-05
fix: add consistent bootc sudo policy

### LINKED ISSUE #142: Consistent sudo policy for `bootc`
I hate that I need to `sudo bootc status`. Anything that doesn't need sudo should "just run" or prompt me to do the priviledged operation.

**Common · issue pipeline**

```text
  ✓  triage     needs kind/ + area/ then /approve
  ·  discussing —
  ▶  queued     —
  ·  claimed    —
  ·  done       —
```

**area:** area/policy &nbsp; **priority:** —  
**maintainer:** —  
**reporter:** comment /claim to fix it yourself, or wait for a contributor

castrojo:
## Summary

- add the existing ublue-polkit-rules bootc sudoers policy to the shared OCI layer
- allow wheel users to run the read/status and update commands without a password prompt

Closes #142

## Validation

- `just check`
- `python3 -m json.tool system_files/shared/etc/containers/policy.json`
- `git diff --check`

`visudo` and `shellcheck` were not run because they are not installed in the checkout environment.


castrojo:
Deferred to the human queue during backlog triage — this is sudo policy on `system_files/shared/`, so it lands on bluefin, bluefin-lts, dakota, and Aurora simultaneously. Recording the review findings so whoever picks it up doesn't redo them.

**The problem is real.** Confirmed on a current image that `bootc status` does require root:

```
$ bootc status --format json
```

**As written, it does not fix this repo's own call site.** `sudoers(5)`:

> If a Cmnd has associated command line arguments, the arguments in the Cmnd must match those given by the user on the command line.

The rule pins `bootc status` with exactly those arguments, but the shipped ujust recipe runs:

```just
# system_files/shared/usr/share/ublue-os/just/default.just:145
$(sudo bootc status --verbose)
```

`--verbose` does not match, so `ujust` still prompts. `sudo bootc upgrade` in `update.just:26` does match exactly and would work.

**Checked and clear:**

- `visudo -cf` → `parsed OK`
- Mode `0644` is acceptable; sudo 1.9.16 loads a 0644 file from `/etc/sudoers.d` in practice, `0440` is convention rather than a hard requirement
- `001-` prefix sorts first, so later drop-ins can still override — correct ordering
- Destructive `bootc install reset --experimental` is correctly not covered

**Open decisions for a maintainer:**

1. Whether to widen to `/usr/bin/bootc status *` so `--verbose` and future flags are covered, versus enumerating exact argument forms.
2. Whether `update` and `upgrade` should be NOPASSWD at all. They mutate the booted OS. Restricting to `status` alone addresses #142's stated complaint without any write-side policy change. The pinned arguments do prevent redirecting to another image, and wheel members can escalate by other means, so the practical delta is the loss of a re-authentication prompt on an OS-mutating command.

Re-derive:

```bash
visudo -cf system_files/shared/etc/sudoers.d/001-bootc
grep -rn 'sudo bootc' system_files/shared/usr/share/ublue-os/just/
man 5 sudoers | grep -A6 'If a Cmnd has associated'
```

hanthor (dismissed):
Requesting changes — the intent is right and the diff is one line, but as written the policy doesn't do what it says. Three concrete problems, all verifiable without a test system.

**1. `bootc update` is not a subcommand.** bootc's verbs are `upgrade`, `switch`, `rollback`, `status`, `edit`, `install`, `container` — there is no `update` (see `bootc.8`, which lists `**bootc upgrade** | Download and queue an updated container image to apply`). That clause is dead weight and grants nothing.

**2. `bootc status` won't match the one place that actually uses it under sudo.** sudoers argument matching is exact — a Cmnd with arguments matches only that exact argument vector. So `/usr/bin/bootc status` permits `sudo bootc status` and nothing else, while the repo's only sudo'd caller is:

```
system_files/shared/usr/share/ublue-os/just/default.just:145
    $(sudo bootc status --verbose)
```

`--verbose` is an extra argument, so that still prompts for a password. The clause as written covers a call site that doesn't exist.

**3. `bootc status` doesn't need root in the first place.** This repo already proves it — `bonedigger-report` calls it unprivileged and treats failure as merely absent data:

```
system_files/bluefin/usr/libexec/bonedigger-report:82
    BOOTC_JSON="$(bootc status --json 2>/dev/null || printf '{}')"
```

So the right fix for `default.just:145` is to drop the `sudo`, not to grant a NOPASSWD rule for it.

---

**Suggested change.** Reduce the drop-in to the one verb that genuinely needs root and is genuinely invoked bare (`update.just:26` is exactly `sudo bootc upgrade`):

```
%wheel ALL=(ALL) NOPASSWD: /usr/bin/bootc upgrade
```

and separately drop the `sudo` from `default.just:145`. If you do want `status` under sudo for some caller I've missed, it needs `/usr/bin/bootc status *` to tolerate flags — but I'd avoid the wildcard and just not sudo a read-only command.

Worth considering whether `/usr/bin/bootc rollback` belongs here too, since it's the natural partner to `upgrade` in a recovery flow and takes no arguments. Your call — out of scope for #142 if you'd rather keep it minimal.

**One packaging check before merge:** sudo ignores drop-ins that are group- or world-writable. Please confirm the OCI layer lands `etc/sudoers.d/001-bootc` as root-owned `0440` (or at least `0644`) — if the build copies `system_files/` with permissive modes, the file is silently skipped and the whole policy is a no-op with no error anywhere. `just check` and the `json.tool`/`git diff --check` validations in the PR body wouldn't catch that. `visudo -cf` on the file would be worth adding to CI given there's no other sudoers coverage in the repo.

castrojo:
@ahmedadan please take a look since we ran into this in dakota today.

ahmedadan (approved):

Danathar (approved):
Approving as the second reviewer. The current one-line diff is exactly what hanthor asked for in his (since dismissed) review: allow wheel users to run `sudo bootc upgrade` without a password — and only that. Because sudo matches arguments exactly, this rule does not cover `bootc switch`, `bootc rollback`, or `bootc upgrade --apply`; just the bare command that `update.just` runs. castrojo confirmed `visudo -cf` parses it and that a 0644 drop-in is honoured. On the policy question (should upgrade be passwordless at all), Jorge answered it on #1134: updating should just work without prompts. ahmedadan has already approved; this is the second.

PR #934, opened 2026-08-05
fix: guard ublue-fastfetch with command -v fastfetch check

### LINKED ISSUE #550: ublue-fastfetch fails with 'exec: fastfetch: not found'
```
DAKOTA 🦖  ·  issue pipeline
─────────────────────────────────────────────────
  ▶  filed      report received
  ·  approved   —
  ·  queued     —
  ·  claimed    —
  ·  done       —
─────────────────────────────────────────────────
  report:       missing     ·  confirms: 0
  area:         —         ·  priority: —
  next action:  same bug? ujust confirm 648
```

---### Before submitting

### ujust report gist URL
*(Not attached)*

### What happened?
When running the helper script `/usr/bin/ublue-fastfetch` (which is copied from upstream `projectbluefin/common` inside the `bluefin/common.bst` BuildStream element), the execution fails with the following error:
```
/usr/bin/ublue-fastfetch: line 25: exec: fastfetch: not found
```
(Or line 27, depending on whether shuffle-logo is enabled or disabled).

This happens because the `fastfetch` binary itself is missing in Dakota. 
`docs/skills/overview.md` documents `fastfetch` under the "Notable Gaps (Priority Order)" table:
```markdown
| Package | Notes |
|---|---|
| fastfetch | System info tool, in both production Bluefins |
```
Since `/usr/bin/ublue-fastfetch` and fish/zsh vendor configurations depend on the presence of `fastfetch`, running terminal sessions or invoking system info commands throws errors.

### Steps to reproduce
1. Run/boot a Dakota image.
2. Open a terminal or shell that executes `/usr/bin/ublue-fastfetch` on startup, or manually run:
   ```bash
   ublue-fastfetch
   ```
3. Observe the command failure: `/usr/bin/ublue-fastfetch: line 25: exec: fastfetch: not found`

### Affected area
GNOME / desktop

### Extra context
We need to add/compile `fastfetch` in Dakota as a BuildStream element or include it as a pre-built binary dependency, so that the underlying tools/scripts in `projectbluefin/common` function correctly.

**Common · issue pipeline**

```text
  ▶  triage     needs kind/ + area/ then /approve
  ·  discussing —
  ·  queued     —
  ·  claimed    —
  ·  done       —
```

**area:** area/bling &nbsp; **priority:** priority/p2  
**maintainer:** set kind/ + area/ labels, then comment /approve  
**reporter:** answer questions; add repro steps if a bug

kylerankin:
## Problem

When `fastfetch` is not installed (e.g., in the Dakota image), the `ublue-fastfetch` wrapper script fails with a confusing error:

```
/usr/bin/ublue-fastfetch: line 25: exec: fastfetch: not found
```

## Fix

Add a `command -v fastfetch` guard at the top of the script. If the binary is missing, the script exits with a clear error message:

```
Error: fastfetch is not installed
```

## Notes

This is a defensive fix in `common`. The consuming repos (bluefin, dakota) must also ensure `fastfetch` is included in their package lists — as documented in `docs/skills/brew-lifecycle.md`, fastfetch "must stay on image regardless" because it's called by ublue-motd before brew runs on first login.

Closes #550


castrojo:
For future reference, let's get used to a world where we'd just accept this and then file an issue to add "install it from homebrew is missing" as an issue - the smaller scoped stuff like this the better.

castrojo:
Thanks for this @kylerankin — you correctly identified the bug and #550 is a real problem.

Closing in favor of #931, which fixes the same issue with slightly different behavior on the missing-binary path:

- **#931 exits `0` silently**; this PR exits `1` and writes to stderr.
- `ublue-fastfetch` is invoked by `ublue-motd` at login. On Dakota, `fastfetch` is intentionally absent, so a non-zero exit with an error message would surface a spurious error on **every login** for those users.
- Silent exit is actually the behavior your own PR description asks for ("harmless no-op"), so this is a small implementation difference, not a disagreement about intent.
- #931 also carries a regression test and documents the guard pattern in `docs/skills/shell-scripts.md` so the next script doesn't repeat it.

Your diagnosis drove the fix — much appreciated, and please do keep them coming.

PR #941, opened 2026-08-07
fix(brew): replace invalid `brew tap --trust` with `brew tap` + `brew trust`

### LINKED ISSUE #814: `brew tap --trust` is invalid syntax — breaks `ujust devmode`, installs on Homebrew 6.0+
### Before submitting

- [ ] I ran ujust report (or am about to — paste the URL below)

### ujust report gist URL

_No response_

## Summary

#665 specifies `brew tap --trust <owner>/<tap>` as the required fix for the Homebrew 6.0 tap-trust requirement, and #695/#728 correctly implemented that spec. The problem is the spec itself: `--trust` is not, and has never been, a valid flag on `brew tap`. The correct interface (confirmed in Homebrew's own docs at https://docs.brew.sh/Tap-Trust) is two separate commands:

```
brew tap <owner>/<tap>
brew trust <owner>/<tap>
```

Every recipe that was "fixed" by #695/#728 is still broken on any current Homebrew 6.0+ install, because `brew tap` rejects the unknown `--trust` flag outright and exits non-zero before doing anything. Since #665 is closed (satisfied by #695), this is a new report rather than a duplicate — the acceptance criteria it was closed against needs correcting, not just re-verifying.

## Environment

- Image: `bluefin-dx:stable`
- `brew --version`: Homebrew 6.0.0 (post 2026-06-11 release)
- Confirmed on a live system as of 2026-07-07, well after both #695 (merged 2026-06-14) and #728 (merged 2026-06-20) shipped.

## Affected locations (as shipped in the image, prior to any local patch)

`/usr/share/ublue-os/just/system.just`:
```
149:        brew tap --trust ublue-os/tap
152:        brew tap --trust ublue-os/experimental-tap
```

`/usr/share/ublue-os/just/apps.just`:
```
12:    brew tap --trust ublue-os/tap
63:    brew tap --trust ublue-os/tap
```

Per #728's own PR description, the same invalid pattern was also introduced/extended in `install-jetbrains-toolbox`, `bbrew`/`cncf` (for `Valkyrie00/homebrew-bbrew`), `install-asus`, and `bazaar-hook`'s `spawn_brew`. I've only directly verified `system.just`/`apps.just` on my own system, but the PR text describes the same `brew tap --trust` construction being applied in those other call sites too.

## Suggested fix

Split every `brew tap --trust <tap>` into the two real commands, keeping the existing silent-fail convention already used elsewhere in these files:

```
brew tap <tap> 2>/dev/null || true
brew trust <tap> 2>/dev/null || true
```

Verified locally: replacing the four lines above in `system.just`/`apps.just` (via `rpm-ostree usroverlay` for a live-tested patch) lets `ujust devmode` complete successfully. Happy to open a PR with this change plus the equivalent fix for `install-jetbrains-toolbox`, `bbrew`/`cncf`, `install-asus`, and `bazaar-hook` if that's useful — let me know which repo/branch convention you'd want it against.

## Why this slipped through twice

The invalid syntax originates in #665's own acceptance criteria, not in the PRs — #695 and #728 both correctly implemented what was specified. Both PRs were AI-assisted (Copilot/Claude) and merged while the CodeRabbit review was still rate-limited on both — no human approval visible on either before merge, and apparently nobody ran the resulting `brew tap --trust` invocation against a real Homebrew 6.0 install before merging either time. Might be worth a CI check that actually shells out to a real `brew tap`/`brew trust` invocation (not just a bats mock accepting any flag) to catch invalid flag/spec combinations like this at PR time, similar to the `brew bundle list` smoke test #728 added for Brewfiles.

### Steps to reproduce

```
$ ujust devmode
...
Usage: brew tap [options] [user/repo] [URL]
...
Error: invalid option: --trust
error: recipe `toggle-devmode` failed with exit code 1
error: recipe `devmode` failed with exit code 1
```

The `brew tap --help` output that gets printed lists the real accepted flags (`--custom-remote`, `--repair`, `-f/--force`, `-d/--debug`, `-q/--quiet`, `-v/--verbose`, `-h/--help`) — no `--trust` among them, which is direct confirmation this isn't an environment quirk, it's invalid syntax on the current `brew tap` command surface.

### Affected area

GNOME / desktop

### Extra context

_No response_

castrojo:
## Problem

`brew tap --trust <owner>/<tap>` is invalid syntax. `--trust` has never been a valid flag on `brew tap`. On Homebrew 6.0+, `brew tap` rejects the unknown flag and exits non-zero **before doing anything**, so `ujust devmode` and several app installers are broken as shipped.

The correct interface ([docs.brew.sh/Tap-Trust](https://docs.brew.sh/Tap-Trust)) is two commands: `brew tap` then `brew trust`.

## Changes

Five call sites fixed — one more than #814 listed:

| File | Recipe |
|---|---|
| `system_files/shared/.../just/apps.just:12` | `install-jetbrains-toolbox` |
| `system_files/shared/.../just/apps.just:63` | `install-asus` (under `set -euo pipefail`) |
| `system_files/bluefin/.../just/system.just:149,152` | `toggle-devmode` |
| `system_files/bluefin/usr/libexec/bazaar-hook:34` | `spawn_brew` — **not in the original report** |

`bbrew`/`cncf` were checked: no `bbrew` recipe exists in `system_files/` anymore, and `cncf` uses `brew bundle` with `trusted: true` Brewfiles, which is already correct.

Docs: `docs/skills/brew-lifecycle.md` actively **taught** the bug in a diff block (`+ brew tap --trust`); reversed and corrected, along with a stale note in `oem-hardware-hooks.md`.

## Tests

`tests/test_brew_tap_trust.bats` — 7 tests. The `brew` mock rejects unknown flags on `tap` exactly as Homebrew 6.0 does, and the real recipes are driven via `just --justfile`.

- Against this fix: **7/7 pass**
- Against pre-fix code: **6/7 fail** — genuine regression coverage, not tests fitted to the change

## Reviewer note — one judgment call

`bazaar-hook` changed `&&` to `;` between tap and install. With `&&`, a user who **already has the tap** gets a non-zero exit from `brew tap` and the cask install silently never runs. This is a behavior change beyond the literal flag fix — say the word if you'd rather keep `&&`.

## Blast radius

`system_files/shared/` → bluefin + bluefin-lts + dakota.

Closes #814

---

## Branch state

Merged `main` at `3de8c769` after #926 landed. The only collision was a
one-line conflict in the `docs/TESTING.md` test table — both PRs append a row.
Resolved by keeping both rows; `Justfile` auto-merged.

Re-verified after that merge: `test_brew_tap_trust.bats` **7/7 pass** and
`test_theming_hook.bats` (from #926) **5/5 pass**.

The 6/7-fail-on-pre-fix-code claim above was re-confirmed independently by
checking out `origin/main`'s `system_files/` under this branch's tests.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -30,6 +30,7 @@ test:
     bats tests/test_hardware_hooks.bats
     bats tests/test_theming_hook.bats
     bats tests/test_nvidia_flatpak_sync.bats
+    bats tests/test_brew_tap_trust.bats
 
 # Preview Bazaar config from this checkout on the local machine
 bazaar-preview:
--- a/Justfile
+++ b/Justfile
@@ -26,6 +26,7 @@ test:
     bats tests/test_bonedigger_report.bats
     bats tests/test_hardware_hooks.bats
     bats tests/test_nvidia_flatpak_sync.bats
+    bats tests/test_brew_tap_trust.bats
 
 # Preview Bazaar config from this checkout on the local machine
 bazaar-preview:
--- a/Justfile
+++ b/Justfile
@@ -25,6 +25,7 @@ test:
     bats tests/test_oem_brew.bats
     bats tests/test_bonedigger_report.bats
     bats tests/test_hardware_hooks.bats
+    bats tests/test_theming_hook.bats
     bats tests/test_nvidia_flatpak_sync.bats
     bats tests/test_brew_tap_trust.bats
 
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -98,6 +98,7 @@ Do not add exemptions for scripts with branching logic.
 | `tests/test_theming_hook.bats` | `10-theming.sh` — Framework/Thelio branches and setup idempotency |
 | `tests/test_brew_preinstall.bats` | Managed Brewfile lifecycle plus user-unit ordering, resource priority, and preset delivery |
 | `tests/test_validate_brewfiles.bats` | Brewfile metadata validation, tap setup failures, ambiguity diagnostics, safe argument passing, and qualified wallpaper/Zed references |
+| `tests/test_brew_tap_trust.bats` | `apps.just`, `system.just`, `bazaar-hook` — `brew tap` + `brew trust` are separate commands; `brew tap --trust` is invalid (#814) |
 
 ## Quality Epic
 
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -89,6 +89,7 @@ Do not add exemptions for scripts with branching logic.
 | `tests/test_bling_fastfetch.bats` | `ublue-bling-fastfetch` — all 9 accent colors, dconf/gsettings fallback chain, FASTFETCH_FORCE_THEME override |
 | `tests/test_changelog.bats` | `changelog.just` — LTS/non-LTS repo selection, URL construction, exit behaviour |
 | `tests/test_ublue_fastfetch.bats` | `ublue-fastfetch` — config reads, shuffle branch, DEFAULT_THEME export to ublue-bling-fastfetch |
+| `tests/test_brew_tap_trust.bats` | `apps.just`, `system.just`, `bazaar-hook` — `brew tap` + `brew trust` are separate commands; `brew tap --trust` is invalid (#814) |
 
 ## Quality Epic
 
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -89,6 +89,7 @@ Do not add exemptions for scripts with branching logic.
 | `tests/test_bling_fastfetch.bats` | `ublue-bling-fastfetch` — all 9 accent colors, dconf/gsettings fallback chain, FASTFETCH_FORCE_THEME override |
 | `tests/test_changelog.bats` | `changelog.just` — LTS/non-LTS repo selection, URL construction, exit behaviour |
 | `tests/test_ublue_fastfetch.bats` | `ublue-fastfetch` — config reads, shuffle branch, DEFAULT_THEME export to ublue-bling-fastfetch |
+| `tests/test_theming_hook.bats` | `10-theming.sh` — Framework/Thelio branches and setup idempotency |
 | `tests/test_brew_tap_trust.bats` | `apps.just`, `system.just`, `bazaar-hook` — `brew tap` + `brew trust` are separate commands; `brew tap --trust` is invalid (#814) |
 
 ## Quality Epic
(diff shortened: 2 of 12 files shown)


clubanderson (commented):
One concern: `|| true` suppresses `brew trust` failures as well as idempotent tap errors. If trust fails, the recipe continues to install from an untrusted tap and hides the actionable error. Prefer guarding only the expected already-tapped case, or verify trust before continuing.

hanthor (approved):
Approving — the core fix is correct and I verified it against real Homebrew rather than from the docs alone. Two notes, one of which answers your reviewer question directly.

**Premise confirmed.** On Homebrew 6.0.15 locally:
- `brew tap --trust` → `Usage: brew tap [options] [user/repo] [URL]`, i.e. a hard usage error before doing anything. The bug is real and exactly as described.
- `brew trust` exists, and bare `brew trust <owner>/<tap>` parses fine without needing `--tap` (matches docs.brew.sh/Tap-Trust). The two-command replacement is right.

Nice catch on the fifth call site in `bazaar-hook` — that one wasn't in #814.

---

**Your judgment call on `&&` → `;` in `bazaar-hook`: I'd keep `&&`.** The rationale in the PR body doesn't hold empirically. Re-tapping an already-tapped repo exits **0**, not non-zero:

```
$ brew tap charmbracelet/tap   # already tapped
$ echo $?
0
```

`brew trust` is idempotent the same way — `Already trusted tap: charmbracelet/tap`, exit 0. So the "user who already has the tap gets a non-zero exit and the cask install silently never runs" failure mode doesn't occur, and `&&` was never unsafe here.

Switching to `;` costs something real: this runs detached in `xdg-terminal-exec`, so if the tap genuinely fails (network, renamed tap), `;` marches on and the user gets a cascade of confusing errors in a terminal that may close before they can read it. `&&` stops at the actual cause. Suggest reverting that hunk to:

```python
f'{brew} tap ublue-os/tap && {brew} trust ublue-os/tap && {brew} install --cask {app}'
```

**Same reasoning for `2>/dev/null || true` in `apps.just` / `system.just`.** Since both commands already exit 0 on the repeat-invocation path, the suppression isn't buying idempotency — it's only hiding genuine failures. In `install-asus` in particular, the recipe runs under `set -euo pipefail` specifically so failures surface, and `|| true` opts these two lines out of that; the user then hits `brew install --cask` against an untrusted tap and gets an error that points at the wrong thing. I'd drop at least the `2>/dev/null` so the real message survives:

```bash
brew tap ublue-os/tap
brew trust ublue-os/tap
```

Neither point is blocking — the PR is a strict improvement over shipped-and-broken either way, and the bats coverage (7/7 pass, 6/7 fail pre-fix) is genuinely load-bearing rather than fitted. Note the mock always exits 0 for a well-formed `brew tap`, so it happens to model the real semantics correctly and doesn't bake in the non-zero assumption.

One housekeeping item: this is currently **DIRTY** — it conflicts with `main` (and with #942/#943, which also touch `Justfile` and `docs/TESTING.md`). Needs a rebase before it can merge regardless of the above.

PR #942, opened 2026-08-07
fix(ujust): use bootc status as source of truth for live image tag

### LINKED ISSUE #820: bctl shows the wrong image tag
### Before submitting

- [ ] I ran ujust report (or am about to — paste the URL below)

### ujust report gist URL

_No response_

### What happened?

bctl reports that I'm on stable/testing, while `bootc status` reports:
```
● Booted image: ghcr.io/projectbluefin/bluefin-lts:stable
        Digest: sha256:5c9c8e050a0c9486f097bbfcbdd1517425814aca56c227bcd5b28d133a44ecc9 (amd64)
       Version: 10 (2026-07-03T22:38:02Z)
```

Image

### Steps to reproduce

run `bctl`.

### Affected area

ujust

### Extra context

This could be related to the reason why `ujust update` fails. This is the only output:
```
⠴ ✗ System Image ━━━━━━━━━ 0/? failed                 0:00:00
```

castrojo:
## Problem

`/usr/share/ublue-os/image-info.json` is baked at **build time** and describes the image the filesystem was *built* as. After a rebase it no longer matches what is booted, so every consumer reports the build-time stream instead of the live one.

Confirmed on this host: the baked file reads `{"image-name":"dakota-gaming","image-tag":"latest",...}` and its `image-ref` carries **no tag at all**.

## Changes

New `system_files/shared/usr/libexec/ublue-image-resolve <image-name|image-tag|image-path>` — bootc-first, baked-file fallback. Consumers updated: `ublue-image-info.sh`, `changelog.just`, `system.just` (`toggle-testing`).

Centralizing avoids three divergent copies of bootc-JSON parsing.

### Schema verification

`bootc status --json` requires root, so the schema was verified from **bootc source** (`crates/lib/src/spec.rs`, v1.16.6): `HostStatus.booted` → `BootEntry.image` → `ImageStatus.image` → `ImageReference.image`. The jq path is `.status.booted.image.image.image`.

Note: the existing `bonedigger-report` uses `.status.booted.imageDigest`, a different field — not a usable precedent for the ref.

### Ref parsing

Splits on the **final path component**, so both `registry:5000/foo/bar:tag` and `repo:tag@sha256:...` parse correctly. This also fixes a latent bug in the old `sed` transport-strip (`s|^[a-z-]+:||` would have eaten `registry:`).

## Output contract — unchanged

`ublue-image-info.sh` still emits `name:tag` plus the lock glyph; verified live (`dakota-gaming:latest 🔓`). Consumers checked: `fastfetch.jsonc`, `tests/test_ublue_image_info.bats`, `docs/TESTING.md`, `docs/contributing/style-guide.md`. Only stderr noise on a missing file/jq changed.

## Tests

`tests/test_ublue_image_resolve.bats` — 23 tests: live-beats-stale, tag+digest, digest-only, ported registry, bootc absent / non-zero / empty / invalid JSON / `booted:null`, transport stripping, jq absent, bad field. Plus an integration test in `test_ublue_image_info.bats`. **43/43 pass** across affected files. Wired into `Justfile`, `unit-tests.yml`, and `docs/TESTING.md`.

## Reviewer notes

- **`image-path` instead of `image-ref`.** bootc refs include a tag; the baked `image-ref` does not. Feeding a bootc ref through the old sed would produce `…/bluefin-lts:stable:testing` in `bootc switch`. `image-path` returns a tagless registry path and centralizes that.
- **Drive-by:** `test_changelog.bats` filtered only `~/.local/bin` for a host `bctl`, but brew ships one too — all 15 tests were failing on `main` locally. Now filters any PATH dir providing `bctl`.
- **Pre-existing, untouched:** 4 failures in `tests/test_update_just.bats` (fail identically on `main`).
- **Follow-up, not in scope:** `bonedigger-report` reads the same baked file and has the same staleness class, affecting bug-repo routing.

## Blast radius

`system_files/shared/` → bluefin + bluefin-lts + dakota.

Closes #820

---

## Branch state

Merged `main` at `d5b56d85` after #926 landed. The only collision was a
one-line conflict in the `docs/TESTING.md` test table — both PRs append a row.
Resolved by keeping both rows.

Re-verified after that merge: `test_ublue_image_resolve.bats` **23/23 pass**
and `test_theming_hook.bats` (from #926) **5/5 pass**.

## Overlap with the sibling PRs in this batch

- **#941** touches `system_files/bluefin/.../just/system.just` at lines 149/152
  (`brew tap`); this PR touches line 249 (`toggle-testing`). Git auto-merges
  those hunks cleanly — verified with a real sequential merge, not `merge-tree`.
- **#943** also edits `docs/skills/shell-scripts.md`, but adds different
  sections (`Bash DEBUG traps…`, `POSIX-sh files…` vs this PR's
  `image-info.json is build-time state…`). No semantic overlap.

Whichever of #941/#942 merges second will need one more `docs/TESTING.md`
resolve. That is inherent to the shared table, not a defect in either branch.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -103,6 +103,9 @@ jobs:
       - name: Run bats (ublue-image-info.sh)
         run: bats tests/test_ublue_image_info.bats
 
+      - name: Run bats (ublue-image-resolve)
+        run: bats tests/test_ublue_image_resolve.bats
+
       - name: Run bats (profile.d — caffeinate.sh + uutils.sh)
         run: bats tests/test_profile_d.bats
 
--- a/Justfile
+++ b/Justfile
@@ -18,6 +18,7 @@ test:
     bats tests/test_ublue_fastfetch.bats
     bats tests/test_ublue_motd.bats
     bats tests/test_ublue_image_info.bats
+    bats tests/test_ublue_image_resolve.bats
     bats tests/test_profile_d.bats
     bats tests/test_dynamic_wallpaper.bats
     bats tests/test_geoclue_latitude.bats
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -62,7 +62,7 @@ Scripts exempt from behavioral testing (shellcheck-only):
 | `usr/share/ublue-os/bling/env.sh` | Sourced helper — sets env vars only |
 | `usr/share/ublue-os/user-setup.hooks.d/20-dynamic-wallpaper.sh` | One-shot hook — logic tested indirectly via setup integration tests |
 | `usr/bin/ublue-motd` | Display-only wrapper — cosmetic tput/glow call, no decision logic |
-| `usr/bin/ublue-image-info.sh` | Read-only reporting wrapper — jq + rpm-ostree status, no branching that affects system state |
+| `usr/bin/ublue-image-info.sh` | Read-only reporting wrapper — delegates resolution to `ublue-image-resolve` (tested separately) plus rpm-ostree status |
 
 **Adding an exemption:** add a row to this table with a one-sentence justification.
 Do not add exemptions for scripts with branching logic.
@@ -89,6 +89,7 @@ Do not add exemptions for scripts with branching logic.
 | `tests/test_bling_fastfetch.bats` | `ublue-bling-fastfetch` — all 9 accent colors, dconf/gsettings fallback chain, FASTFETCH_FORCE_THEME override |
 | `tests/test_changelog.bats` | `changelog.just` — LTS/non-LTS repo selection, URL construction, exit behaviour |
 | `tests/test_ublue_fastfetch.bats` | `ublue-fastfetch` — config reads, shuffle branch, DEFAULT_THEME export to ublue-bling-fastfetch |
+| `tests/test_ublue_image_resolve.bats` | `ublue-image-resolve` — bootc-status-first tag/name/ref resolution, digest and ported-registry refs, image-info.json fallback |
 
 ## Quality Epic
 
--- a/docs/skills/shell-scripts.md
+++ b/docs/skills/shell-scripts.md
@@ -153,6 +153,30 @@ Apply to any script reading system files via stdin redirect.
 
 ---
 
+### image-info.json is build-time state — use `ublue-image-resolve` for live identity
+
+`/usr/share/ublue-os/image-info.json` is baked when the image is built, so it
+describes what the filesystem was *built* as, not what is booted. After a rebase
+it lies (see [#820](https://github.com/projectbluefin/common/issues/820)).
+
+Ask `/usr/libexec/ublue-image-resolve` instead. It reads
+`.status.booted.image.image.image` from `bootc status --json` and falls back to
+the baked file when bootc is absent, non-zero, or reports no booted image.
+
+```bash
+IMAGE_RESOLVE="${IMAGE_RESOLVE:-/usr/libexec/ublue-image-resolve}"
+IMAGE_TAG="$("${IMAGE_RESOLVE}" image-tag)"   # stable
+IMAGE_NAME="$("${IMAGE_RESOLVE}" image-name)" # bluefin-lts
+IMAGE_PATH="$("${IMAGE_RESOLVE}" image-path)" # ghcr.io/projectbluefin/bluefin-lts
+```
+
+Never split a ref on the first `:` — registry ports (`registry:5000/foo:tag`)
+and digests (`repo:tag@sha256:...`) both break that. The resolver handles both.
+
+In bats, mock `bootc` onto PATH and point `IMAGE_RESOLVE` at the real script.
+
+---
+
 ### Assert env-var export against the subshell consumer, not exec
 
 `exec` inherits all shell variables whether exported or not — asserting `DEFAULT_THEME`
--- a/system_files/bluefin/usr/share/ublue-os/just/changelog.just
+++ b/system_files/bluefin/usr/share/ublue-os/just/changelog.just
@@ -7,12 +7,13 @@ changelogs:
         exec bctl changelogs
     fi
 
-    # Get Image Tag
-    IMAGE_INFO_FILE="${IMAGE_INFO_FILE:-/usr/share/ublue-os/image-info.json}"
-    TAG="$(jq -r '.["image-tag"]' < "${IMAGE_INFO_FILE}")"
+    # Get Image Tag from the live system (bootc status), not the build-time
+    # image-info.json, which goes stale after a rebase.
+    IMAGE_RESOLVE="${IMAGE_RESOLVE:-/usr/libexec/ublue-image-resolve}"
+    TAG="$("${IMAGE_RESOLVE}" image-tag || true)"
 
     # Select the correct upstream repo based on image stream
-    IMAGE_NAME="$(jq -r '.["image-name"] // empty' < "${IMAGE_INFO_FILE}")"
+    IMAGE_NAME="$("${IMAGE_RESOLVE}" image-name || true)"
     if [[ "$IMAGE_NAME" == "dakota" ]]; then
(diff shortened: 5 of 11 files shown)


hanthor (approved):
Approving — the diagnosis is right, the centralization is the correct shape, and I traced the ref-parsing logic case by case rather than trusting the description. Three notes, one of which I'd like confirmed before this lands.

**Parsing verified.** I walked `ref_path` / `ref_basename` / `strip_transport` through the awkward inputs:

| input | result |
|---|---|
| `registry:5000/foo/bar:tag` | path `registry:5000/foo/bar`, tag `tag` ✅ |
| `registry:5000/foo/bar` (no tag) | path unchanged, port intact ✅ |
| `repo:tag@sha256:…` | digest stripped before split ✅ |
| `bluefin:latest` (no slash) | path `bluefin`, tag `latest` ✅ |
| `ostree-image-signed:docker://ghcr.io/…` | `##*://` handles it ✅ |
| `ostree-unverified-registry:ghcr.io/foo` | transport regex strips it ✅ |

The `[^0-9]` guard in `strip_transport` is what keeps `registry:5000/…` from being eaten — that's a subtle case and it's handled correctly. Agreed this is a real improvement on the old `s|^[a-z-]+:||`.

---

**1. Please confirm `bootc status --json` works unprivileged — the whole fix rests on it.**

The PR body says it "requires root". If that's true on a real host, the headline benefit doesn't reach the most visible consumer: `fastfetch.jsonc:23` calls `/usr/bin/ublue-image-info.sh` in the **user's** shell, so `bootc status` would fail there, silently fall through to the baked file, and post-rebase fastfetch would keep reporting the stale stream — exactly the bug being fixed.

For what it's worth, `bootc-status.8` documents no root requirement and explicitly recommends `bootc status --json` as *the* way to programmatically detect a bootc deployment, which suggests it is unprivileged-safe. But since you verified the schema from source specifically because you couldn't run it, this is the one thing I can't confirm from here. A single `bootc status --json` as a normal user on a Bluefin box settles it.

**2. Two resolver invocations means two `bootc status --json` per shell.**

```bash
IMAGE_NAME="$("${IMAGE_RESOLVE}" image-name ...)"
IMAGE_TAG="$("${IMAGE_RESOLVE}" image-tag ...)"
```

Each call independently runs `bootc status --json` and pipes through `jq`. Since fastfetch runs this on every new terminal, that's 2× bootc + 2× jq per shell on top of the existing `rpm-ostree status --booted`. Worth collapsing — either a combined mode (`ublue-image-resolve image-name-tag` emitting `name:tag`), or have the resolver cache the JSON in an env var the caller can pass through. Not blocking, but it's cheap to fix now and annoying to notice later as "terminals got slower".

**3. Minor: half-resolved output.**

```bash
if [[ -n "${IMAGE_NAME}" || -n "${IMAGE_TAG}" ]]; then
	echo -n "${IMAGE_NAME}:${IMAGE_TAG}"
```

`||` means one empty side still prints, yielding `bluefin:` or `:stable`. Reachable with a digest-pinned bootc ref on a host with no baked file (tag unresolvable, name fine). `&&` — or emitting just the non-empty part — would be tidier.

---

Good call flagging `bonedigger-report` as the same staleness class rather than scope-creeping into it, and the `test_changelog.bats` `bctl` drive-by is a real fix (brew ships one, so that filter was wrong for anyone with brew — which is everyone on Bluefin).

Housekeeping: currently **DIRTY**. Your branch-state note anticipates the `docs/TESTING.md` collision with #941, so that's understood — just needs the resolve.

hanthor (changes requested):
## Superseded by #1020, and the parts that aren't would regress `main`

The stated goal — "use bootc status as source of truth for live image tag" — already landed:

```
$ git log --oneline -1 -- system_files/shared/usr/bin/ublue-image-info.sh
b68f6ad fix(report): read booted tag from bootc, not baked image-info (#1020)
```

`main`'s `ublue-image-info.sh` now reads:

```bash
BOOTED_REF="$(bootc status --json 2>/dev/null | \
	jq -r '.status.booted.image.image.image // .spec.image.image // empty' 2>/dev/null)"
case "${BOOTED_REF}" in
	*@*) ;; # digest-pinned ref, no tag to extract
	*:*) IMAGE_TAG="${BOOTED_REF##*:}" ;;
esac
```

That is this PR's premise, already shipped, and it handles the digest-pinned case that the `ublue-image-resolve` approach here does not visibly cover.

The branch is `dirty` — it conflicts with `main` in five places, and each conflict is a place where `main` has moved past this PR rather than a mechanical clash:

```
$ git merge pr942
CONFLICT (content): Merge conflict in docs/TESTING.md
CONFLICT (modify/delete): docs/skills/shell-scripts.md deleted in HEAD and modified in pr942.
CONFLICT (content): Merge conflict in system_files/bluefin/usr/share/ublue-os/just/changelog.just
CONFLICT (content): Merge conflict in system_files/shared/usr/bin/ublue-image-info.sh
CONFLICT (content): Merge conflict in tests/test_changelog.bats
```

Two of those are regressions, not conflicts to resolve in this PR's favour:

**1. `changelog.just` — reintroduces a hardcoded repo table that `main` deliberately extracted.** `main` now delegates the name/tag → repo grammar to a shipped helper:

```bash
# Select the correct upstream repo based on image stream. The name/tag ->
# repo grammar is owned by /usr/libexec/ublue-image-repo; do not restate it.
REPO="$("${UBLUE_IMAGE_REPO_BIN:-/usr/libexec/ublue-image-repo}" \
    --default "projectbluefin/bluefin" "$IMAGE_NAME" "$TAG")"
```

`/usr/libexec/ublue-image-repo` exists on `main` (`ls system_files/shared/usr/libexec/` → `bootc-update-stage  brew-preinstall  ublue-image-repo`). This PR replaces that call with an inline `if/elif` chain for `dakota` / `bluefin-lts` / default — exactly the restatement the comment on `main` warns against. That grammar is consumed by three images, so a second copy is a real divergence risk.

**2. `docs/skills/shell-scripts.md` was deleted on `main`** (`b151bbe docs(skills): migrate oversized skills to per-skill directories, drop exemptions`). This PR adds 24 lines to it, resurrecting a file the docs layout has moved away from.

Also worth flagging: this PR edits `.github/workflows/unit-tests.yml`, which widens the review surface beyond the stated fix.

### What I'd suggest

Close this as superseded by #1020, and if `ublue-image-resolve` still buys something #1020 doesn't (a single resolver shared between `ublue-image-info.sh` and `changelog.just`), open a fresh PR against current `main` that layers it **on top of** `/usr/libexec/ublue-image-repo` rather than replacing it, and leaves `docs/skills/shell-scripts.md` deleted.

---

hanthor:
Closing — the premise has already shipped via #1020.

```
$ git log -1 --format='%h %s' -- system_files/shared/usr/bin/ublue-image-info.sh
b68f6ad fix(report): read booted tag from bootc, not baked image-info (#1020)
```

`main` already reads the booted tag from `bootc status` rather than the build-time `image-info.json`, which is exactly the problem this PR set out to solve. The diagnosis here was right — it just got fixed by another route while this sat.

Worth flagging why rebasing this isn't the cheap option: its conflict resolution would actively undo two things `main` has since done deliberately.

- **`changelog.just`** — this branch reinstates a hardcoded dakota/lts `if`/`elif` chain, but `main` extracted that into `/usr/libexec/ublue-image-repo` with an explicit "do not restate it" note. Rebasing would put the restatement back.
- **`docs/skills/shell-scripts.md`** — deleted on `main` by `b151bbe`. This branch still modifies it, so the merge reads as a resurrection rather than an edit.

So the rebase isn't a mechanical replay; it's re-litigating two settled decisions to land a fix that's already in.

The one piece here that may not be covered by #1020 is the `image-path` helper (tagless registry path, to stop `bootc switch` producing `…/bluefin-lts:stable:testing`). If that is still missing on `main`, it's worth a small focused PR on current `main` rather than reviving this branch — and the reasoning in this description is the right starting point for it.

Reopen if #1020 turns out not to cover your case.

---

PR #943, opened 2026-08-07
fix(bling): rearm bash-preexec DEBUG trap on each prompt

### LINKED ISSUE #869: bling: atuin history silently stops recording in bash — bash-preexec 0.6.0 + Fedora array PROMPT_COMMAND
### Before submitting

### ujust report gist URL

error: justfile does not contain recipe `report`

### What happened?

**DISCLAIMER**: I was annoyed by this bug for a few weeks, and could not figure it out by myself, so I asked GitHub Copilot to help me find a solution. In the end, we landed in the possible actual bug, but with a practical workaround. This was _not_ an unattended run. I verified everything myself. I also wrote this report myself.

`atuin` does not save the bash history when `bling.sh` is enabled. This has already been reported before, but I took various steps to reproduce and find a workaround.

This report sounds like a duplicate, but it's a follow-up on [ublue-os/bluefin#4081](https://github.com/ublue-os/bluefin/issues/4081) because I can't re-open that bug, and I feel like the report actually belongs here, since it's an issue related with [bling.sh](https://github.com/projectbluefin/common/blob/main/system_files/shared/usr/share/ublue-os/bling/bling.sh)

The root cause seems to be a bug already fixed in [rcaloras/bash-preexec#188](https://github.com/rcaloras/bash-preexec/issues/188), however just downloading from `master/HEAD` doesn't entirely fixes it, as I am hit by another bug as reported in [rcaloras/bash-preexec#186](https://github.com/rcaloras/bash-preexec/issues/186) (still not solved).

Basically the problem is that a combination of factors cause `tap -p DEBUG` to land in the wrong place, so it prevents `atuin` from loading (notice it's missing from here).
```
❯ declare -p PROMPT_COMMAND
declare -a PROMPT_COMMAND=([0]=$'__bp_precmd_invoke_cmd\n_mise_hook_prompt_command' [1]="_direnv_hook" [2]=$'starship_precmd\n__bp_trap_string="$(trap -p DEBUG)"\ntrap - DEBUG\n__bp_install' [3]="__systemd_osc_context_precmdline" [4]="__vte_precmd" [5]="__vte_osc7" [6]="__zoxide_hook" [7]="__bp_interactive_mode")
```

### Steps to reproduce

1. `ujust bluefin-cli` to enable `bling`.
2. As per [this dosu doc](https://app.dosu.dev/e3630b91-3a35-46b9-a8d3-b0c1b3ef6331/documents/670ff1e2-7e5c-4591-bc3e-fdac3c87e6d9) add the following line to `~/.bashrc` after the `bling.sh` line:
```
[ "$(command -v atuin)" ] && eval "$(atuin init bash --disable-up-arrow)"
```
3.  Open a _new_ terminal, run a few commands, then run `trap -p DEBUG`, it will show up empty. `CTRL+R` shows `atuin`, which proves it's enabled, but the commands are not there. Hitting the up arrow shows the commands, so `bash` itself is recording them.
4. Run `declare -p PROMPT_COMMAND` and see the same output as above.

### Affected area

brew / CLI tools

### Extra context

Since `ujust report` is not working, here's some information about my system:
```
❯ rpm-ostree status
State: idle
AutomaticUpdates: stage; rpm-ostreed-automatic.timer: inactive
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/bluefin-dx:stable
                   Digest: sha256:70adf7d21c68810e580ff3ed908380632091e2d51504e55feaf88a65fc42b81c
                  Version: 44.20260721 (2026-07-21T01:58:58Z)
❯ brew list --versions bash-preexec atuin
atuin 18.18.1
bash-preexec 0.6.0
```

I tried multiple ways to fix this, and the only thing that really worked was to enter in developer mode (`sudo bootc usr-overlay`) then append this to the end of `/usr/share/ublue-os/bling/bling.sh`:
```
if [ "${BLING_SHELL}" = "bash" ] && type __bp_preexec_invoke_exec >/dev/null 2>&1; then
    __bling_rearm_debug_trap() { trap '__bp_preexec_invoke_exec "$_"' DEBUG; }
    PROMPT_COMMAND+=(__bling_rearm_debug_trap)
fi
```

Since I can't keep my system running that way, I rebooted (which disabled developer mode), then I added this to my `~/.bashrc` as a temporary workaround (after `bling` and `atuin`).
```
__atuin_rearm_debug_trap() { trap '__bp_preexec_invoke_exec "$_"' DEBUG; }
if [[ "$(type -t __bp_preexec_invoke_exec)" == function ]]; then
    PROMPT_COMMAND+=(__atuin_rearm_debug_trap)
fi
```

This is the prompt after the patch (notice atuin is now there, even though the trap debug is still in the wrong place):
```
❯ declare -p PROMPT_COMMAND
declare -a PROMPT_COMMAND=([0]=$'__bp_precmd_invoke_cmd\n_mise_hook_prompt_command' [1]="_direnv_hook" [2]=$'starship_precmd\n__bp_trap_string="$(trap -p DEBUG)"\ntrap - DEBUG\n__bp_install' [3]="__systemd_osc_context_precmdline" [4]="__vte_precmd" [5]="__vte_osc7" [6]="__zoxide_hook" [7]="__atuin_rearm_debug_trap" [8]="__bp_interactive_mode")
```

So it works, for now, and Bluefin can choose to patch `bling.sh`, or wait for the upstream `bash-preexec` fix, but at least I felt like sharing my workaround here, in case someone else faces the same problem.

I hope this is helpful.

castrojo:
## Root cause — confirmed against bash-preexec 0.6.0 source

Verified by reading the tagged 0.6.0 source directly, not from memory:

- L74: the installer is deferred by appending a string to `PROMPT_COMMAND`:
  `__bp_trap_string="$(trap -p DEBUG)"; trap - DEBUG; __bp_install`
- L289-293: `__bp_install()` **returns early** if `"${PROMPT_COMMAND[*]}"` already contains `__bp_precmd_invoke_cmd`
- The self-removal code reads `existing_prompt_command="${PROMPT_COMMAND:-}"` and later assigns `PROMPT_COMMAND='__bp_precmd_invoke_cmd'`

On Fedora, `PROMPT_COMMAND` is an **array** (bash ≥ 5.1). `${PROMPT_COMMAND}` and bare assignment address **element `[0]` only**. So once any other hook (mise, direnv, starship, vte, systemd, zoxide) pushes the installer out of element `[0]`, it is **never removed** — and runs on every prompt. Its first act is `trap - DEBUG`; `__bp_install` then returns early. From prompt 2 onward the DEBUG trap is permanently empty.

That is exactly the reporter's `declare -p` output: `__bp_precmd_invoke_cmd` in `[0]`, installer stranded in `[2]`, atuin's hook absent.

Reproduced empirically in a real prompt loop: `cycle1 trap: [trap -- '__bp_preexec_invoke_exec "$_"' DEBUG]`, `cycle2: []`, `cycle3: []`. With this fix, all three cycles retain the trap.

Matches upstream rcaloras/bash-preexec#188 (fixed upstream but insufficient alone) and #186 (still open).

## Non-obvious discovery

Bash does not inherit the DEBUG trap into functions without `functrace`: `trap -p DEBUG` **inside a function always returns empty**, and `trap - DEBUG` inside a function is a no-op — but `trap ... DEBUG` inside a function *does* set it globally.

Two consequences: a detector-guard is impossible (hence the unconditional re-arm), and it invalidated the first simulation, which passed **vacuously**. Written up in `docs/skills/shell-scripts.md`.

## Changes

- New `system_files/shared/.../bling/bash-preexec-rearm.sh` — re-arms the exact trap bash-preexec installs, once per prompt
- `bling.sh` sources it **last**, under the bash guard, so the hook is the final `PROMPT_COMMAND` entry

`bling.sh` stays POSIX (it is under an sh-dialect shellcheck gate); the bash-only logic lives in the sibling file. `BLING_DIR` allows bats to point at the repo tree.

No vendored copy of bash-preexec.

## Tests — 14 cases, all pass

Tests **1 and 4 are negative controls** that pin the bash-preexec stand-in to the real defect, so the positive cases cannot pass vacuously.

Covered: array `PROMPT_COMMAND` trap survival, atuin-style hook still firing, idempotency on double-source, scalar `PROMPT_COMMAND` (no regression), bash-preexec absent, readonly `PROMPT_COMMAND`, missing helper, zsh source-safety.

Also green: `bash -n` / `sh -n`, both CI shellcheck invocations, `just check`, `pre-commit run --all-files`.

## Confidence and limits

High that the DEBUG trap now survives — the failure and its repair are both directly reproduced. Two things only real hardware can prove:

1. atuin actually writing records end-to-end (tests use a preexec stub, not atuin 18.18.1)
2. that no downstream hook on a real session appends its own `trap - DEBUG` **after** our entry

Ordering is safe for everything bling itself initializes.

## Reviewer note

The re-install is **unconditional** per prompt, forced by the functrace scoping above; it matches the reporter's validated workaround. A third party's own raw DEBUG trap set after bling would be overwritten each prompt — but such a trap already breaks bash-preexec today.

## Blast radius — highest in this batch

`bling.sh` is sourced by **every interactive bash shell** on bluefin, bluefin-lts, and dakota. Recommend lab validation before merge.

Closes #869

---

## Branch state

Updated onto `main` after #926 landed (via `gh pr update-branch`). No conflict —
this PR's `docs/TESTING.md` row and `Justfile` entry merged cleanly.

## Overlap with the sibling PRs in this batch

**#942** also edits `docs/skills/shell-scripts.md`, but adds different sections
(this PR: `Bash DEBUG traps are invisible inside functions`, `POSIX-sh files
cannot hold bash array code`; #942: `image-info.json is build-time state`).
Verified non-conflicting with a real sequential merge.

No file overlap with #941.

## Pre-merge checklist for the reviewer

- [ ] Lab-validate on real hardware — the tests use a preexec stub, not atuin
      18.18.1, so end-to-end recording is unproven
- [ ] Confirm the unconditional per-prompt trap re-install is acceptable
      (see reviewer note above; a detector-guard is impossible due to bash's
      functrace scoping)

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -9,6 +9,7 @@ test:
     bats tests/test_privileged_setup.bats
     bats tests/test_bling.bats
     bats tests/test_bling_sh.bats
+    bats tests/test_bling_preexec_rearm.bats
     bats tests/test_luks_tpm2.bats
     bats tests/test_rechunker_group_fix.bats
     bats tests/test_bling_fastfetch.bats
--- a/Justfile
+++ b/Justfile
@@ -26,6 +26,7 @@ test:
     bats tests/test_oem_brew.bats
     bats tests/test_bonedigger_report.bats
     bats tests/test_hardware_hooks.bats
+    bats tests/test_theming_hook.bats
     bats tests/test_nvidia_flatpak_sync.bats
 
 # Preview Bazaar config from this checkout on the local machine
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -89,6 +89,7 @@ Do not add exemptions for scripts with branching logic.
 | `tests/test_setup_scripts.bats` | `ublue-system-setup`, `ublue-user-setup` — hook runner logic |
 | `tests/test_privileged_setup.bats` | `ublue-privileged-setup` — privileged hook runner logic |
 | `tests/test_bling.bats` | `ublue-bling` — shell config injection install/uninstall |
+| `tests/test_bling_preexec_rearm.bats` | `bling/bash-preexec-rearm.sh` — DEBUG trap re-arm with array/scalar `PROMPT_COMMAND`, idempotency, degradation when bash-preexec is absent |
 | `tests/test_luks_tpm2.bats` | `luks-tpm2-autounlock` — UUID parsing, device resolution, cryptenroll flag construction |
 | `tests/test_rechunker_group_fix.bats` | `rechunker-group-fix` — group/gshadow append, duplicate detection, format |
 | `tests/test_bling_fastfetch.bats` | `ublue-bling-fastfetch` — all 9 accent colors, dconf/gsettings fallback chain, FASTFETCH_FORCE_THEME override |
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -90,6 +90,7 @@ Do not add exemptions for scripts with branching logic.
 | `tests/test_bling_fastfetch.bats` | `ublue-bling-fastfetch` — all 9 accent colors, dconf/gsettings fallback chain, FASTFETCH_FORCE_THEME override |
 | `tests/test_changelog.bats` | `changelog.just` — LTS/non-LTS repo selection, URL construction, exit behaviour |
 | `tests/test_ublue_fastfetch.bats` | `ublue-fastfetch` — config reads, shuffle branch, DEFAULT_THEME export to ublue-bling-fastfetch |
+| `tests/test_theming_hook.bats` | `10-theming.sh` — Framework/Thelio branches and setup idempotency |
 
 ## Quality Epic
 
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -84,6 +84,7 @@ Do not add exemptions for scripts with branching logic.
 | `tests/test_setup_scripts.bats` | `ublue-system-setup`, `ublue-user-setup` — hook runner logic |
 | `tests/test_privileged_setup.bats` | `ublue-privileged-setup` — privileged hook runner logic |
 | `tests/test_bling.bats` | `ublue-bling` — shell config injection install/uninstall |
+| `tests/test_bling_preexec_rearm.bats` | `bling/bash-preexec-rearm.sh` — DEBUG trap re-arm with array/scalar `PROMPT_COMMAND`, idempotency, degradation when bash-preexec is absent |
 | `tests/test_luks_tpm2.bats` | `luks-tpm2-autounlock` — UUID parsing, device resolution, cryptenroll flag construction |
 | `tests/test_rechunker_group_fix.bats` | `rechunker-group-fix` — group/gshadow append, duplicate detection, format |
 | `tests/test_bling_fastfetch.bats` | `ublue-bling-fastfetch` — all 9 accent colors, dconf/gsettings fallback chain, FASTFETCH_FORCE_THEME override |
--- a/docs/skills/index.json
+++ b/docs/skills/index.json
@@ -592,7 +592,7 @@
       ],
       "description": "Shell script authoring and testability. Use when writing or testing shell scripts under system_files/, removing scripts, or adding bats tests.",
       "version": "1.0",
-      "last_updated": "2026-08-01",
+      "last_updated": "2026-08-06",
       "doc_type": "reference"
     },
     {
--- a/docs/skills/shell-scripts.md
+++ b/docs/skills/shell-scripts.md
@@ -1,7 +1,7 @@
 ---
 name: shell-scripts
 version: "1.0"
-last_updated: "2026-08-01"
+last_updated: "2026-08-06"
 id: shell-scripts
 one_line_purpose: Write and test shell scripts under system_files/.
 entry_point: docs/skills/shell-scripts.md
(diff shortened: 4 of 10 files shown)


hanthor (approved):
Approving. I independently reproduced every load-bearing claim in this PR rather than taking the writeup on trust, and all five hold exactly.

**The functrace discovery — confirmed on bash 5.3.15:**

```
trap -p DEBUG inside a function   → []              (empty, as claimed)
trap -  DEBUG inside a function   → no-op           (outer trap survives)
trap ... DEBUG inside a function  → sets globally   (as claimed)
```

So a detector-guard genuinely is impossible and the unconditional re-arm is forced, not lazy. Good call writing that up in `docs/skills/shell-scripts.md` — that's the kind of thing that costs the next person an afternoon.

**The PROMPT_COMMAND array asymmetry — confirmed against bash-preexec 0.6.0 source:**

```bash
# L291  — array-aware
if [[ "${PROMPT_COMMAND[*]:-}" == *"__bp_precmd_invoke_cmd"* ]]; then return 1; fi
# L324  — scalar: element [0] ONLY
existing_prompt_command="${PROMPT_COMMAND:-}"
# L336  — scalar assignment: clobbers [0], leaves [1..n] untouched
PROMPT_COMMAND='__bp_precmd_invoke_cmd'
```

That asymmetry *is* the bug, and it's sharper than the description: the early-return check is array-aware while the self-removal is not, so once the installer is pushed past element `[0]` it becomes permanently self-sustaining — L291 keeps short-circuiting `__bp_install` while the stranded `trap - DEBUG` fires every prompt. And `${PROMPT_COMMAND}` on an array does expand to `[0]` alone (verified: `PROMPT_COMMAND=(a b c)` → `[a]`).

The re-armed string `'__bp_preexec_invoke_exec "$_"'` matches bash-preexec L295 character for character, and single-quoting correctly defers `$_` to trap time.

**Implementation details I checked:**
- `(unset PROMPT_COMMAND) 2>/dev/null` is a sound readonly probe — `unset` on a readonly var fails even in a subshell.
- `PROMPT_COMMAND+=('...')` on a *scalar* correctly promotes to an array (`declare -a PROMPT_COMMAND=([0]="orig" [1]="new")`), so the ≥5.1 branch is safe even when something set it as a plain string.
- The 5.1 version gate is right — array `PROMPT_COMMAND` is exactly a 5.1 feature.
- Idempotency guard uses `${PROMPT_COMMAND[*]-}`, array-aware. Correct, and notably not the mistake bash-preexec itself made.

---

**One durability concern, non-blocking.** The trap string is hardcoded to bash-preexec 0.6.0's internal. If Fedora ships a version that changes it, this re-arms a *stale* trap every prompt and the failure is silent in exactly the same way as the original bug — atuin loads, CTRL+R works, nothing records. Worth a guard: assert at build/test time that the literal matches the `trap ... DEBUG` line in the shipped `bash-preexec.sh`, so a package bump trips CI instead of user history. A single grep in the bats suite would cover it.

Agreed on the blast radius — `bling.sh` on every interactive bash shell across three images warrants lab validation before merge, and your two "only real hardware can prove" caveats are the right two.

Housekeeping: **DIRTY** — conflicts with `main`/siblings on `Justfile` + `docs/TESTING.md`, needs a rebase.

PR #966, opened 2026-08-08
feat(ujust): honor ACTION in toggle-updates for non-interactive use

hanthor:
## What

`ujust toggle-updates` declares an `ACTION="prompt"` parameter but the recipe body never reads it: it always opens the `bctl --screen updates` GUI panel (when bluefinctl is available) or blocks on an interactive `gum choose` prompt. Nothing can drive it from a script, an automation, or an SSH test step.

This PR makes the recipe read the parameter (via just's `{{ ACTION }}` interpolation — shebang recipe bodies receive parameters only through interpolation, never as positional args):

- `ujust toggle-updates enable` / `disable` / `cancel` (case-insensitive) select the action directly and skip **both** the bctl panel and the gum prompt.
- Any other value — including the default `ACTION="prompt"` — keeps the existing interactive behavior byte-for-byte.

## Why

This is the `projectbluefin/common` contract change required by [projectbluefin/testsuite#499](https://github.com/projectbluefin/testsuite/issues/499) so the `toggle-updates` scenario can be tested non-interactively. The follow-up testsuite scenario (projectbluefin/testsuite PR, `@requires_toggle_action`) drives `ujust toggle-updates enable|disable` over SSH and restores the timer state.

## Tests

- `tests/test_update_just.bats` extended with ACTION-path coverage: `ACTION=enable` enables `uupd.timer` without prompting, `ACTION=disable` disables the fallback timer, `ACTION=cancel` exits without touching the timer, and `ACTION` is case-insensitive. All 16 tests pass locally (bats 1.11.1).
- `just --unstable --fmt --check` passes on `update.just`.
- Verified end-to-end with mocked `systemctl`/`sudo`/`gum`: `just toggle-updates enable` runs `systemctl enable uupd.timer` and prints `Updates have been enabled` without consulting gum; the default `prompt` path still consults gum.

## Notes

- The positional contract is `ujust toggle-updates <action>`; `ujust toggle-updates ACTION=enable` is not the documented form (just passes the literal `ACTION=enable` string positionally for shebang recipes).
- The test harness (`_extract_script`) now substitutes the just placeholder `{{ ACTION }}` with `${ACTION:-prompt}` so the extracted script reads the ACTION environment variable, matching how the recipe behaves under `just`.

Closes the prerequisite for projectbluefin/testsuite#499.


castrojo (approved):
LGTM. Reviewed the recipe logic against the current `toggle-updates`:

- `ACTION_VALUE="{{ ACTION }}"` + `case "${ACTION_VALUE,,}"` correctly routes `enable`/`disable`/`cancel` and the default `"prompt"` falls into `*)`, preserving the existing bctl-panel → gum-prompt path byte-for-byte.
- Post-`esac` timer selection (`uupd.timer` vs `rpm-ostreed-automatic.timer`) and the `Enable`/`Disable` branches work unchanged because the non-interactive arms set the same `SELECTED_OPTION` values gum would return.
- `cancel` exiting 0 before touching the timer is correct for the testsuite restore flow (projectbluefin/testsuite#499).
- bats coverage is solid: enable/disable/cancel plus case-insensitivity, and the `_extract_script` sed substitution of `{{ ACTION }}` → `${ACTION:-prompt}` faithfully mirrors just's interpolation for shebang recipes.

Two nits (non-blocking): the new comment block has a stray blank line splitting it mid-sentence, and "Pass ... as ACTION or a positional argument" is slightly misleading since `ujust toggle-updates ACTION=enable` is *not* the supported form (positional only, as the PR body correctly notes).

CI: the `validate` failure is unrelated flake — `scripts/check-oci-refs.py` died with `http.client.RemoteDisconnected` talking to GHCR. All build/test/E2E-adjacent checks pass; a re-run should go green.

PR #1035, opened 2026-08-28
fix(flatpak): correct metered check in appstream refresh service

### LINKED ISSUE #1027: `flatpak-appstream-refresh.service` checks the wrong NetworkManager metered values
### What happened, or what do you want?

NetworkManager defines the values as:

| Value | Enum | Meaning | Current unit behavior |
|---:|---|---|---|
| 0 | `NM_METERED_UNKNOWN` | Unknown | Runs |
| 1 | `NM_METERED_YES` | Explicitly metered | Skips |
| 2 | `NM_METERED_NO` | Explicitly not metered | Skips (wrong) |
| 3 | `NM_METERED_GUESS_YES` | Guessed metered | Runs (wrong) |
| 4 | `NM_METERED_GUESS_NO` | Guessed not metered | Runs |

The values are defined in NetworkManager's authoritative
[`NMMetered` API documentation](https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/nm-dbus-types.html#NMMetered).
That documentation also says most applications should treat
`NM_METERED_GUESS_YES` like `NM_METERED_YES`.

Expected behavior: skip values `1` and `3`, and allow explicitly or guessed
unmetered values `2` and `4`. Keeping the current fail-open behavior for unknown
value `0` is a separate policy choice.

Actual behavior: the service skips explicitly unmetered value `2`, while it can
download AppStream metadata on guessed-metered value `3`.

AI disclosure: this bug report was created by OpenAI Codex, an AI coding agent,
after inspecting the installed unit, current upstream source, and authoritative
NetworkManager documentation. The human reporter reviewed it before submission.

### Reproduction or evidence

The current condition's truth table can be reproduced without changing network
settings or downloading metadata:

```sh
for v in 0 1 2 3 4; do
  if [ "$v" != "1" ] && [ "$v" != "2" ]; then
    printf '%s: runs\n' "$v"
  else
    printf '%s: skips\n' "$v"
  fi
done
```

It prints:

```text
0: runs
1: skips
2: skips
3: runs
4: runs
```

The affected unit is still present on `main` at commit
[`99bc5d1`](https://github.com/projectbluefin/common/commit/99bc5d1c658b68e31a8c4e92fd345a52214c8171):

- [`flatpak-appstream-refresh.service`, line 7](https://github.com/projectbluefin/common/blob/99bc5d1c658b68e31a8c4e92fd345a52214c8171/system_files/shared/usr/lib/systemd/system/flatpak-appstream-refresh.service#L7)
- [The vendor preset enables the service](https://github.com/projectbluefin/common/blob/99bc5d1c658b68e31a8c4e92fd345a52214c8171/system_files/shared/usr/lib/systemd/system-preset/02-flatpak-appstream-refresh.preset#L1)

The mapping is not ambiguous: the same official NetworkManager documentation
defines `1` as explicitly metered, `2` as explicitly not metered, `3` as guessed
metered, and `4` as guessed not metered.

### Image and version

_No response_

## Image and version

I'm on a custom image, but the faulty condition is identical locally and
on current upstream `main`.

## Extra context

A minimal correction preserving the current treatment of unknown value `0` is:

```ini
ExecCondition=bash -c 'v=$(busctl get-property org.freedesktop.NetworkManager /org/freedesktop/NetworkManager org.freedesktop.NetworkManager Metered 2>/dev/null | cut -d" " -f2); [ "$v" != "1" ] && [ "$v" != "3" ]'
```

The condition could also use a `case` statement to make the enum names/intent
clearer and reduce the chance of another numeric-value mistake.

HuntedRaven7:
ExecCondition skipped NM_METERED_YES(1) and NM_METERED_NO(2), so an explicitly-unmetered connection 2 was never refreshed while a guessed-metered connection 3 was. NetworkManager's NMMetered docs say to treat NM_METERED_GUESS_YES(3) as metered and all else as unmetered.

Skip only 1 and 3; keep 0/unknown fail-open. Moved to a case statement for clearer intent and to avoid future numeric miscomparisons.

Closes [#1027](https://github.com/projectbluefin/common/issues/1027)

Assisted-by: big-pickle (opencode) via OpenCode

### DIFF
--- a/system_files/shared/usr/lib/systemd/system/flatpak-appstream-refresh.service
+++ b/system_files/shared/usr/lib/systemd/system/flatpak-appstream-refresh.service
@@ -4,7 +4,7 @@ Documentation=https://docs.projectbluefin.io
 After=network-online.target flatpak-system-helper.service
 Wants=network-online.target
 ConditionPathExists=/usr/bin/flatpak
-ExecCondition=bash -c 'v=$(busctl get-property org.freedesktop.NetworkManager /org/freedesktop/NetworkManager org.freedesktop.NetworkManager Metered 2>/dev/null | cut -d" " -f2); [ "$v" != "1" ] && [ "$v" != "2" ]'
+ExecCondition=bash -c 'v=$(busctl get-property org.freedesktop.NetworkManager /org/freedesktop/NetworkManager org.freedesktop.NetworkManager Metered 2>/dev/null | cut -d" " -f2); case "$v" in 1|3) exit 1;; esac'
 StartLimitIntervalSec=600
 StartLimitBurst=3
 


hanthor (approved):
Correct fix, and the bug was real: the old condition `[ "$v" != "1" ] && [ "$v" != "2" ]` skipped the refresh for both `NM_METERED_YES(1)` *and* `NM_METERED_NO(2)` — meaning an explicitly-unmetered connection was being denied a refresh it should always get, while `NM_METERED_GUESS_YES(3)` (which NetworkManager's own docs say to treat as metered) fell through and refreshed anyway. The new `case "$v" in 1|3) exit 1;; esac` skips only on 1 and 3 and fail-opens (runs the refresh) for everything else, including unknown/empty `$v` on a `busctl` failure — matches NM's documented semantics exactly. CI green (`test`, `Compose PR test image`, `validate`, both arch builds).

---

PR #348, opened 2026-05-24
fix(#266): drop monospace-font-name override — use GNOME default (Adwaita Mono)

hanthor:
## Summary

- Removes `monospace-font-name="JetBrains Mono 11"` from the gschema override, letting GNOME fall back to its upstream default (Adwaita Mono)

## Context

Issue #266 asked to drop our custom font diff now that GNOME has a text size slider. The blocker was whether to replace JetBrains Mono with a Nerd Font-patched Adwaita Mono. As noted in the issue:

- `che/nerd-fonts` COPR has not yet packaged Adwaita Mono ([discussion](https://discussion.fedoraproject.org/t/che-nerd-fonts/87161/18))
- Downloading from nerd-fonts.com is blocked in the build process (PR #271 hit this)
- The upstream Adwaita Mono renders nerd font icons slightly differently but is not broken

The cleanest path forward: drop the override entirely. GNOME will use Adwaita Mono by default. JetBrains Mono Nerd Font remains available opt-in via `fonts.Brewfile` for users who want it.

Closes #266

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1039, opened 2026-08-29
feat(security): provide shared runsc provisioning for Bluefin-family hosts

### LINKED ISSUE #1038: feat(security): provide shared runsc provisioning for Bluefin-family hosts
## Outcome

Provide one shared, trusted Bluefin-family host provisioning path for gVisor's
`runsc` OCI runtime so Review and other isolated workloads can select
`--runtime=runsc` without each image repository maintaining its own installer.

## Ownership

Common's `system_files/shared/` is distributed to the Bluefin family and owns
reusable host infrastructure. The initial implementation was developed in
projectbluefin/bluefin#1142; this issue re-homes that work rather than creating
a second implementation.

## Required behavior

Preserve the #1142 contract: a pinned upstream release; exact x86_64 and
aarch64 identities; unsupported-architecture rejection before download;
HTTPS-only acquisition; digest verification before archive inspection or
extraction; expected-member allowlisting; the required `gvisor-bin` payload;
safe install/update/remove ownership and symlink protection; atomic
publication; idempotency; `/usr/local/bin/runsc` discovery; unchanged Podman
default runtime; no Review packaging; no `ignore-cgroups=true`; and no
host-network workaround.

## Consumers and acceptance

- Originating requirement: projectbluefin/bluefin#1139
- Superseded implementation location: projectbluefin/bluefin#1142
- Consumer security contract: projectbluefin/review#348 and merged #349
- Future Review execution engine: projectbluefin/review#362

Common provisioning alone does not prove native acceptance. After a Bluefin
Testing image consumes this change, prove `ujust runsc install`, version
reporting, rootless `podman --runtime=runsc`, `OCIRuntime=runsc`, ordinary
networking, lifecycle behavior, no cgroup or host-network workaround, the
Review fail-closed path, and separate arm64 acceptance.

joshyorko:
## Summary

- re-home the reviewed runsc provisioner into Common's shared Bluefin-family overlay
- expose `ujust runsc install`, `ujust runsc update`, and `ujust runsc remove` from Common's canonical shared recipe
- preserve the pinned release, architecture/digest checks, archive allowlist, ownership protection, atomic lifecycle, and no-runtime-fallback contract from Bluefin #1142

Closes #1038
Refs projectbluefin/bluefin#1139
Supersedes the implementation location in projectbluefin/bluefin#1142
Consumer contract: projectbluefin/review#348 and merged #349
Future consumer: projectbluefin/review#362

## Why Common

Common owns reusable host infrastructure in `system_files/shared/`, which is
distributed through the Bluefin-family image composition path. Bluefin consumes
Common's image and copies its shared overlay into the resulting image. The
runsc capability is not GNOME-specific or image-specific.

## Validation

- `just check`
- `shellcheck -S warning system_files/shared/usr/libexec/bluefin-runsc`
- `npx --yes bats tests/test_runsc_provisioning.bats` — 13/13
- affected Python tests with ephemeral pytest/jsonschema dependencies — 109 passed
- ephemeral `uv run ... pre-commit run --all-files` — passed
- `git diff --check`
- Bluefin composition source proof: Common is consumed as `COMMON_IMAGE`, and
  Bluefin copies Common's shared overlay; old Bluefin-local helper and
  `60-custom.just` are absent from testing

The host's broad BATS sweep has unrelated existing setup/changelog/update
failures; all 13 transferred runsc cases pass. No local image build or native
runtime proof is claimed.

## Native acceptance boundary

After a capable Bluefin Testing image consumes this Common layer, native
acceptance must separately prove `ujust runsc install`, `runsc --version`,
rootless `podman --runtime=runsc`, `OCIRuntime=runsc`, ordinary networking,
install/update/remove lifecycle, no `ignore-cgroups=true`, no host networking,
the real Review fail-closed path, and separate arm64 evidence. This PR does
not close Bluefin #1139 or Review #348.

### DIFF
--- a/docs/skills/shell-scripts/SKILL.md
+++ b/docs/skills/shell-scripts/SKILL.md
@@ -86,6 +86,38 @@ Quick reference — full patterns with WRONG/CORRECT examples in
 Standard file layout, mocking, and pitfalls in
 [references/bats-patterns.md](references/bats-patterns.md).
 
+## Bluefin-family host runsc provisioning
+
+Common's shared overlay owns the reusable host-side gVisor runtime provisioner
+for Bluefin-family images. The provisioner does not change Podman's default
+runtime; consumers explicitly select `runsc` when they need the outer
+isolation boundary.
+
+The pinned upstream release is `release-20260817.0`. The helper selects the
+exact architecture asset and digest before it inspects or extracts the archive:
+
+- `x86_64`: `ae345a8c1466586b3a163fb534301913da663a97b8ed446bc711b2e1963a32c5`
+- `aarch64`: `a3c2443e9564dbf500893e66fd2463be3b79fe42f66825971c44dc1624d454b2`
+
+The acquisition is HTTPS-only and the archive member allowlist requires both
+`runsc` and its adjacent `gvisor-bin` payload. Installation stages a complete
+version, publishes `/usr/local/bin/runsc` atomically, and records ownership so
+install/update/remove refuse foreign, unowned, or symlinked paths. Repeating
+install or update is idempotent, and a failed update leaves the active version
+in place. The supported user commands are:
+
+```text
+ujust runsc install
+ujust runsc update
+ujust runsc remove
+```
+
+This capability does not add `ignore-cgroups=true`, use host networking, or
+claim native runtime acceptance. A consumer must separately prove executable
+`runsc`, rootless `podman --runtime=runsc`, `OCIRuntime=runsc`, ordinary
+networking, lifecycle behavior, and each supported architecture on real
+hardware.
+
 ## Shellcheck Reference
 
 Directive syntax, SC code notes, and quoting fix examples in
--- a/system_files/shared/usr/libexec/bluefin-runsc
+++ b/system_files/shared/usr/libexec/bluefin-runsc
@@ -1,0 +1,253 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+RUNSC_VERSION="release-20260817.0"
+RUNSC_ROOT="/usr/local/libexec/bluefin-runsc"
+RUNSC_LINK="/usr/local/bin/runsc"
+RUNSC_MARKER="${RUNSC_ROOT}/.bluefin-owned"
+RUNSC_MARKER_CONTENT="bluefin-runsc ownership marker v1"
+RUNSC_WORK_DIR=""
+RUNSC_STAGE=""
+RUNSC_LINK_DIR=""
+RUNSC_INSTALL_STAGE=""
+RUNSC_RELEASE_DIR=""
+RUNSC_RELEASE_PUBLISHED=0
+RUNSC_ROOT_CREATED=0
+
+die() {
+    echo "bluefin-runsc: $*" >&2
+    return 1
+}
+
+usage() {
+    echo "Usage: bluefin-runsc {install|update|remove}" >&2
+    return 2
+}
+
+cleanup() {
+    if [[ -n "${RUNSC_LINK_DIR}" && -d "${RUNSC_LINK_DIR}" ]]; then
+        rm -rf -- "${RUNSC_LINK_DIR}"
+    fi
+    if [[ -n "${RUNSC_STAGE}" && -d "${RUNSC_STAGE}" ]]; then
+        rm -rf -- "${RUNSC_STAGE}"
+    fi
+    if [[ -n "${RUNSC_INSTALL_STAGE}" && -d "${RUNSC_INSTALL_STAGE}" ]]; then
+        rm -rf -- "${RUNSC_INSTALL_STAGE}"
+    fi
+    if [[ "${RUNSC_RELEASE_PUBLISHED}" -eq 0 && -n "${RUNSC_RELEASE_DIR}" \
+        && -d "${RUNSC_RELEASE_DIR}" && ! -L "${RUNSC_RELEASE_DIR}" ]]; then
+        rm -rf -- "${RUNSC_RELEASE_DIR}"
+    fi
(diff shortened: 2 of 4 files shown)


hanthor (approved):
Read the full `bluefin-runsc` helper (253 lines) and the 301-line bats suite, not just the summary. This is a genuinely careful implementation for a security-sensitive, `sudo`-invoked installer:

- **Supply chain**: per-arch pinned SHA256, HTTPS-only + TLS1.2+ + `--proto '=https'`, hash verified *before* `tar -tjf` ever parses the archive, then an explicit member allowlist checked before extraction (`runsc`, `gvisor-bin/*`, `containerd-shim-runsc-v1` — anything else aborts). Extraction uses `--no-same-owner --no-same-permissions` plus explicit `chmod`, so the archive can't smuggle setuid bits or odd ownership.
- **Directory-confusion safety**: every root/release/link path is checked for `-L` (symlink) before being trusted, and an ownership marker (`cmp`-verified fixed content) distinguishes installer-owned directories from pre-existing/foreign ones — `remove` and `install` both refuse to touch anything not carrying that marker. This is exactly the right defense against a symlink planted at `/usr/local/libexec/bluefin-runsc` before a `sudo` invocation.
- **Atomicity**: release publish is `mv -T` (rename, not copy-over), and the `/usr/local/bin/runsc` symlink is replaced via stage-then-`mv -Tf`, avoiding a window where the link is missing or half-written. A `trap cleanup EXIT` rolls back staged/partial state on any failure, verified by the bats case that fails an `update` mid-`mv` and confirms the previously-active release and link are untouched.
- **Scope discipline**: the recipe (`shared.just`) only calls the helper — no `ignore-cgroups`, no host networking, no default-runtime change — and one bats test greps the helper + recipe to pin exactly that (`runsc recipe and helper do not change defaults or weaken runtime isolation`).

I did not independently re-derive the pinned SHA256 values against the real upstream `google/gvisor` release artifact (that needs fetching a real multi-MB release asset, out of scope for a review), so that trust boundary rests on the author's stated verification. The PR is explicit and correctly scoped about what it does *not* prove — native `podman --runtime=runsc` acceptance on real hardware, arm64 evidence, the Review fail-closed path — and defers those to a separate acceptance step rather than claiming them here.

CI green: `validate`, `test`, `Compose PR test image`, and both arch builds all pass; `Trivy` is neutral (non-blocking) and the manifest/E2E jobs are correctly skipped on a PR.

---

joshyorko:
@hanthor Independently verified the real `release-20260817.0` artifacts. Fresh SHA256 recomputation matches both the upstream `SHA256SUMS` asset and GitHub's reported asset digests:

- x86_64: `ae345a8c1466586b3a163fb534301913da663a97b8ed446bc711b2e1963a32c5`
- aarch64: `a3c2443e9564dbf500893e66fd2463be3b79fe42f66825971c44dc1624d454b2`

The focused provisioning suite passes 13/13 and shellcheck passes. Native Podman/runsc acceptance, arm64 host evidence, and Review fail-closed validation remain separate follow-up gates.

joshyorko:
lgtm

castrojo (changes requested):
## Review: blocking — the installer targets a read-only path on bootc

Thanks for this — gVisor support is genuinely wanted. But as written the helper cannot work on any Bluefin host, and because it lands in `system_files/shared/` it would ship broken to bluefin, bluefin-lts, dakota **and** Aurora simultaneously.

### 1. `/usr/local` is read-only on bootc (blocking)

`system_files/shared/usr/libexec/bluefin-runsc:6-7`:

```sh
RUNSC_ROOT="/usr/local/libexec/bluefin-runsc"
RUNSC_LINK="/usr/local/bin/runsc"
```

Both are written at runtime (lines 63, 72, 211, 226, 231). Verified on a live Bluefin host:

```
$ grep ' / ' /proc/mounts
composefs:... / overlay ro,relatime,...,verity=require

$ ls -ld /usr/local
drwxr-xr-x 3 root root 64 Nov 11  2011 /usr/local     # ostree epoch ts = deployment content
```

`/` is mounted `ro` with `verity=require`, and `/usr/local` is part of that verity-protected tree — it is *not* the usual distro symlink to `/var/usrlocal`. Every `install -d`/`mv`/`ln` in the helper will fail `EROFS`. This matches our own contract in `docs/skills/bootc.md`: "`/usr/` — read-only on the running system; bootc-managed".

**Fix:** put the versioned payload under `/var` (persistent, never reset by bootc) and expose it via a path that doesn't mutate `/usr`.

### 2. Published release directory is `0700` (blocking)

`bluefin-runsc:211-215` — `mktemp -d` creates the staging dir mode `0700`, and it's moved into the published release tree without a mode change. Only `runsc` and `gvisor-bin/*` get chmod'd; the containing directory does not. A rootless Podman process resolving `runsc` can't traverse it and gets `EACCES` — which defeats the rootless contract asserted in `docs/skills/shell-scripts/SKILL.md:117`.

**Fix:** publish the release dir `0755`, and add a test that execs the installed link as a *different* non-root user (the current tests run as the creating user and only assert `-x`, so they can't catch this).

### 3. The 13 new tests never run (should fix before merge)

`tests/test_runsc_provisioning.bats` is not referenced by `Justfile` or `.github/workflows/unit-tests.yml` — both enumerate BATS files explicitly:

```
$ grep -rn 'runsc' Justfile .github/workflows/unit-tests.yml
(no matches)
```

So the green Unit Tests check on this PR did not execute any of the added tests. The extensionless helper is likewise missing from the ShellCheck list.

**Fix:** register the file in `just test` and in `unit-tests.yml`, and add `bluefin-runsc` to the shellcheck gate.

---

Happy to re-review once (1) and (2) are addressed — (1) is the one that has to change shape, the rest are mechanical. 🙏

castrojo:
Ok better idea, it's 2026 we need this thing let's just add it at the image layer, this is just duct tape.

PR #1042, opened 2026-08-29
refactor(just): move 00-entry.just to shared with optional imports

Siddhj2206:
## Summary

Moves `system_files/bluefin/usr/share/ublue-os/just/00-entry.just` to
`system_files/shared/usr/share/ublue-os/just/00-entry.just` and makes **every**
import in the entrypoint optional (`import?`):

- `apps.just`, `default.just`, `shared.just`, `update.just` — now optional
- `changelog.just`, `system.just` — now optional (were mandatory, Bluefin-only)
- `flutter.just`, `60-custom.just`, `60-bonedigger.just` — already optional

## Why

The `00-entry.just` dispatcher is consumed by every variant via the shared
`ujust` wrapper (`system_files/shared/usr/bin/ujust` → `just --justfile
/usr/share/ublue-os/just/00-entry.just`), so it belongs in `shared/` rather
than `bluefin/`. With all-optional imports, the shared dispatcher degrades
gracefully: any variant ships exactly the recipe set it has, and `_default`
always runs.

## Impact (verified)

- **bluefin**: no break — `Containerfile:47-48` merges both `system_files`
  trees recursively; final `/usr/share/ublue-os/just/` tree is identical and
  all 9 files exist (`import?` loads when present).
- **bluefin-lts**: same merge pattern as bluefin, unaffected.
- **dakota**: `elements/bluefin/common.bst:74-77` copies both trees; entrypoint
  now arrives via the shared copy. Dakota overlays
  `default/changelog/system/flutter.just` via `just-overrides.bst` as before.
- **aurora**: does not consume `projectbluefin/common` (pins
  `ghcr.io/get-aurora-dev/common` — its own entrypoint is unchanged), zero
  effect.
- No repo docs, tests, or workflows hardcode the old repo path — `ujust`
  wrapper and `test_ujust.bats` reference the runtime path only.

## Verification

- [x] `just check` passes (entrypoint now found under `shared/`)
- [x] `git mv` rename detected + all imports `import?`
- [ ] CI on this PR

Closes N/A — no issue tracked.

### DIFF
--- a/system_files/shared/usr/share/ublue-os/just/00-entry.just
+++ b/system_files/shared/usr/share/ublue-os/just/00-entry.just
@@ -8,12 +8,12 @@ _default:
 
 # Imports
 
-import "/usr/share/ublue-os/just/apps.just"
+import? "/usr/share/ublue-os/just/apps.just"
 import? "/usr/share/ublue-os/just/changelog.just"
-import "/usr/share/ublue-os/just/default.just"
-import "/usr/share/ublue-os/just/shared.just"
+import? "/usr/share/ublue-os/just/default.just"
+import? "/usr/share/ublue-os/just/shared.just"
 import? "/usr/share/ublue-os/just/system.just"
-import "/usr/share/ublue-os/just/update.just"
+import? "/usr/share/ublue-os/just/update.just"
 import? "/usr/share/ublue-os/just/flutter.just"
 import? "/usr/share/ublue-os/just/60-custom.just"
 import? "/usr/share/ublue-os/just/60-bonedigger.just"
--- a/system_files/shared/usr/share/ublue-os/just/00-entry.just
+++ b/system_files/shared/usr/share/ublue-os/just/00-entry.just
@@ -9,10 +9,10 @@ _default:
 # Imports
 
 import "/usr/share/ublue-os/just/apps.just"
-import "/usr/share/ublue-os/just/changelog.just"
+import? "/usr/share/ublue-os/just/changelog.just"
 import "/usr/share/ublue-os/just/default.just"
 import "/usr/share/ublue-os/just/shared.just"
-import "/usr/share/ublue-os/just/system.just"
+import? "/usr/share/ublue-os/just/system.just"
 import "/usr/share/ublue-os/just/update.just"
 import? "/usr/share/ublue-os/just/flutter.just"
 import? "/usr/share/ublue-os/just/60-custom.just"


HuntedRaven7:
Verified the diff and every consumer wired to this entrypoint — LGTM with one suggestion.

What the diff actually is (confirmed): a git mv of 00-entry.just from bluefin/ to shared/ (auto-detected rename, 2 ins / 2 del) plus relaxing import → import? for changelog.just and system.just. Nothing else touched.
Claims checked against the tree:

- ujust wrapper (system_files/shared/usr/bin/ujust:38) and test_ujust.bats reference only the runtime path /usr/share/ublue-os/just/00-entry.just, so the repo-path move can't affect them.

- bluefin / bluefin-lts: both merge common /system_files/shared + /system_files/bluefin into the same target (bluefin Containerfile:47-48, bluefin-lts Containerfile:17-18); neither ships its own 00-entry.just. Both still deliver changelog.just/system.just via the bluefin tree, so import? behaves exactly like import — final runtime just-tree is byte-identical.

- dakota (projectbluefin/dakota): elements/bluefin/common.bst:74-77 copies both trees and files/just-overrides/ ships its own changelog.just/system.just — same result; entrypoint now correctly arrives via shared/.

- aurora: pins ghcr.io/get-aurora-dev/common, zero impact.

- No references to the old repo path anywhere in projectbluefin/common.

Suggestion (not blocking): the new behavior — a variant without changelog.just/system.just still getting a working ujust — is currently unexercised. Every existing consumer provides both files, so import? is a pure no-op today. A cheap smoke test (run just -f .../shared/00-entry.just --list against a tree missing the two files via the existing test_ujust.bats harness, or a one-off in unit-tests.yml) would lock in the change's actual reason to exist.

Minor nit: PR body says "Containerfile:47-48" for bluefin, but the merge lines it cites are in the bluefin repo's Containerfile, not this one — the line number set is right, just label the repo.

hanthor (commented):
Why do this just to system and changelog? Seems like this is a specific solution for your custom image not something that is general purpose.

system_files/shared/usr/share/ublue-os/just/00-entry.just
hanthor, line 1: maybe make them all `import?` I don't see why not

Siddhj2206: we can do that. For now, I only made files that come from bluefin be `import?`

Do I make them all `import?`

hanthor (approved):
Read the current diff directly: it's a `git mv` of `00-entry.just` from `bluefin/` to `shared/` plus relaxing all six previously-mandatory imports to `import?`. This already reflects the "make them all `import?`" suggestion from the review thread — that conversation is resolved and the code matches it.

Verified the reasoning holds: the `ujust` wrapper and `test_ujust.bats` reference only the runtime path, not the repo path, so the move itself is safe. Both bluefin and bluefin-lts merge `shared/` + `bluefin/` into the same target and currently ship both `changelog.just`/`system.just` from the bluefin tree, so `import?` is a behavior-preserving no-op for every existing consumer today — it only changes what happens for a *future* variant missing those files (graceful skip instead of hard `just` failure), which is the actual point of the change.

No CI has run yet (0 check runs) — likely pending a maintainer's workflow-run approval. Given the change is a rename plus a strictness relaxation with no logic touched, I'm comfortable approving on the verified diff alone; worth confirming CI comes back clean before merge regardless.

---

PR #1109, opened 2026-09-12
fix(brew): preserve user-owned preinstall packages

### LINKED ISSUE #915: brew: brew-preinstall adopts user-installed packages into the managed set
What: `system_files/shared/usr/libexec/brew-preinstall` writes every package declared in `preinstall.d/*.Brewfile` into its managed state after `brew bundle`, without first checking whether the user had already installed it themselves.

`brew bundle` is idempotent, so an already-installed package is a no-op — but it still lands in the `packages`/`casks` arrays of `~/.local/share/ublue-os/brew-preinstall-state.json`. If the OS later drops that line from a Brewfile, the diff marks it "dropped from managed set" and it is uninstalled out from under the user.

Concrete case: a user runs `brew install htop`. A later image adds `htop` to `system-cli.Brewfile`, so it becomes managed. A still later image removes it, and the user loses a package they installed themselves and never asked us to manage.

The script is otherwise careful about this boundary — it explicitly checks `brew list --formula` / `brew list --cask` before removing, and its comments state it "never touches packages the user added outside our managed set". That guarantee holds only for packages we never declared, not for ones we transiently declared.

Fix: snapshot installed formulae and casks *before* the `brew bundle` pass. When writing state, keep everything previously managed, but only adopt newly declared entries that were not already installed. Packages the user got to first stay theirs.

Found during adversarial review of #852/#853. Deliberately not fixed there to keep that PR to one logical change — the cask lifecycle work was already reviewed and approved on its current semantics, and this changes the ownership model for formulae too.

Automatable: yes (logic change plus bats coverage in `tests/test_brew_preinstall.bats`)

Danathar:
## Summary

Fixes #915.

## Root cause

`brew-preinstall` previously wrote every formula and cask declared by the
current managed Brewfiles into `brew-preinstall-state.json` after a successful
`brew bundle`. Because `brew bundle` is idempotent, a package already installed
by the user was indistinguishable from one installed by Bluefin. If the
declaration was later removed, the persisted entry authorized Bluefin to
uninstall the user's package.

## What changed

- Snapshot installed formulae and casks immediately before the bundle pass.
- Preserve declarations already present in the previous managed state.
- Add a declaration to managed state only when it was absent from the
  pre-bundle inventory and the bundle pass succeeds.
- Fail closed without bundling or writing state if either inventory query fails.
- Continue using the existing basename-aware matching for tap-qualified names,
  the existing hash fast path, cask lifecycle, ChairLift handoff, atomic state
  writes, and legacy state files without a `casks` key.

Formulae and casks now follow the same ownership rule: previously user-owned
items remain outside Bluefin's managed state and therefore remain installed
when their temporary Brewfile declaration disappears. Items installed by the
managed reconciliation remain removable, and previously managed items remain
managed across later reconciliations.

## Regression coverage

Added BATS coverage for:

- user-owned formula and cask adoption followed by declaration removal;
- Bluefin-owned formula and cask installation followed by removal;
- previously managed formulae and casks remaining managed;
- legacy formula state compatibility;
- fail-closed behavior when pre-bundle inventory cannot be collected.

The new ownership tests fail against the pre-fix implementation: both the
formula and cask user-owned cases were adopted into state.

## Validation

- `/var/tmp/bats-core/bin/bats tests/test_brew_preinstall.bats` — 56/56 pass.
- `bash -n system_files/shared/usr/libexec/brew-preinstall` — pass.
- `shellcheck -S warning system_files/shared/usr/libexec/brew-preinstall` — pass.
- `just check` — pass.
- `python3 scripts/generate_skill_index.py --check` — pass.
- `PATH=/var/tmp/bluefin-common-validation/bin:$PATH pre-commit run --all-files` — pass.
- `PATH=/var/tmp/bluefin-common-validation/bin:/var/tmp/bats-core/bin:$PATH just test` — all 109 Python tests pass; the suite then reaches 5 existing `tests/test_ujust.bats` command-availability failures. The same failures reproduce on `origin/main`.
- Full BATS suite excluding the documented libvirt integration test — 426/433 pass; the 7 failures are the same baseline failures in `tests/test_shared_just.bats` (2) and `tests/test_ujust.bats` (5), with no `brew-preinstall` failures.
- `git diff --check` — pass.


castrojo:
@Danathar: This branch is conflicted on the generated  files. Since  is true, or you can rebase against  and regenerate with . Once rebased, this is ready for maintainer review.

castrojo (approved):
LGTM. Rebased cleanly onto main, regenerated skill index, and verified all 60 tests in test_brew_preinstall.bats pass.

castrojo:
Rebased cleanly onto main and resolved the docs/skills/index conflict. I have approved; needs a second maintainer review (@hanthor, @ahmedadan, or @repires) to satisfy +2 approvals before landing.

ahmedadan (approved):

hanthor (approved):

PR #1112, opened 2026-09-12
feat(countme): report to countme.projectbluefin.io from the images

castrojo:
## Why the counts are empty

No image reports to the first-party service:

| image | service | target |
| --- | --- | --- |
| Bluefin | standard Fedora repo config | Fedora |
| Bluefin LTS | `bluefin-lts-countme.service` → `dnf makecache` | Fedora/EPEL mirrors |
| Dakota | `/usr/libexec/dakota-countme` | `mirrors.fedoraproject.org` |

Dakota's timer already carries `Documentation=https://countme.projectbluefin.io`, but the script it triggers sends to Fedora.

The service has been accepting and recording pings the whole time — a probe to `/metalink` lands in `telemetry_events` correctly. Nothing was calling it. `telemetry_events` holds 30 rows total, across every image, ever.

## What this adds

A shared weekly reporter for every image consuming `system_files/shared`:

- `/usr/libexec/bluefin-countme` — sends image name, tag, flavor, arch, and a coarse age bucket. No machine-id, hostname, token, or persistent identifier.
- `bluefin-countme.service` — `DynamicUser=yes`, `StateDirectory=bluefin-countme`, `ProtectSystem=strict`, address families limited to INET/INET6.
- `bluefin-countme.timer` — `OnBootSec=15m`, `OnUnitInactiveSec=3d`, `AccuracySec=1h`, `RandomizedDelaySec=12h`. No `Persistent=`: it only applies to `OnCalendar=` timers, and `OnBootSec=15m` already covers systems that were powered off.
- `03-bluefin-countme.preset` — enables the timer.

## Details that matter

**The tag is resolved through bootc.** The baked `image-tag` is the compose-time tag (`latest`) and promotion retags the same digest, so the tag a system actually follows only exists in bootc. Same resolution order as `/usr/bin/ublue-image-info.sh`. Without this every system reports `latest` — which is exactly what the live data shows today.

**Unusable metadata is dropped, not sent.** A value that is empty, `null`, or still contains an unexpanded `${...}` is left out of the query. The live dataset already has a row whose tag is the literal string `${IMAGE_TAG}${IMAGE_FLAVOR}${ARCH}`; that is indistinguishable from a real deployment and cannot be counted. If `image-name` itself is unusable the script reports nothing rather than guessing, since a guess lands in the wrong image's population.

**The 7-day throttle is load-bearing.** The server stores no identifier and *cannot* deduplicate, so the one-ping-per-system-per-week guarantee has to hold client-side. The timer deliberately fires more often than weekly so a machine that was powered off does not miss its window.

**A failed ping does not update `lastrun`.** It retries next firing. A missed week is a gap in the data; a week wrongly marked as reported is a wrong number.

**Opt-out is checked before any state is written**, so opting out leaves no trace on disk. Honours `/etc/projectbluefin/countme/disabled` and the legacy `/etc/dakota-countme/disabled`.

Upstream Fedora and EPEL reporting is untouched and runs alongside this.

## Verification

Run against the live service with a probe image name, then removed from D1:

```
$ bluefin-countme
reported __probe-lts-hwe__ bucket 1
$ bluefin-countme          # immediately again
already reported 0d ago, skipping
```

The row landed with `repo=__probe-lts-hwe__ tag=stable flavor=main arch=x86_64 bucket=1`, and was deleted afterwards.

Guard rails:

```
$ IMAGE_INFO_FILE=bad.json bluefin-countme     # {"image-name":"${IMAGE_NAME}"}
no usable image-name in bad.json, not reporting
$ IMAGE_INFO_FILE=missing.json bluefin-countme
no usable image-name in missing.json, not reporting
```

`shellcheck` clean, `bash -n` clean.

## Follow-ups, not in this PR

- `projectbluefin/dakota` builds from its own `files/countme/` and does not consume `system_files/shared`, so `dakota-countme` still points at Fedora and needs the same change.
- `projectbluefin/bluefin-lts` keeps `dnf makecache` for upstream EPEL reporting; whether it also picks up this shared unit depends on whether it consumes `system_files/shared`.

### DIFF
--- a/system_files/shared/usr/lib/systemd/system-preset/03-bluefin-countme.preset
+++ b/system_files/shared/usr/lib/systemd/system-preset/03-bluefin-countme.preset
@@ -1,0 +1,1 @@
+enable bluefin-countme.timer
--- a/system_files/shared/usr/lib/systemd/system/bluefin-countme.service
+++ b/system_files/shared/usr/lib/systemd/system/bluefin-countme.service
@@ -1,0 +1,22 @@
+[Unit]
+Description=First-party Bluefin CountMe reporting
+Documentation=https://countme.projectbluefin.io
+Wants=network-online.target
+After=network-online.target
+ConditionPathExists=/run/ostree-booted
+ConditionPathExists=!/etc/projectbluefin/countme/disabled
+ConditionPathExists=!/etc/dakota-countme/disabled
+
+[Service]
+Type=oneshot
+DynamicUser=yes
+StateDirectory=bluefin-countme
+ExecStart=/usr/libexec/bluefin-countme
+
+# Nothing identifying is sent, and nothing outside its own state is read.
+PrivateTmp=yes
+ProtectSystem=strict
+ProtectHome=yes
+NoNewPrivileges=yes
+RestrictAddressFamilies=AF_INET AF_INET6
+SystemCallFilter=@system-service
--- a/system_files/shared/usr/lib/systemd/system/bluefin-countme.timer
+++ b/system_files/shared/usr/lib/systemd/system/bluefin-countme.timer
@@ -1,0 +1,16 @@
+[Unit]
+Description=Weekly first-party Bluefin CountMe reporting
+Documentation=https://countme.projectbluefin.io
+
+[Timer]
+# Fires more often than weekly so a system that was off does not miss its week.
+# The client throttles to one ping per 7 days, which is what keeps the count
+# honest: the server stores no identifier and cannot deduplicate.
+OnBootSec=15m
+OnUnitInactiveSec=3d
+AccuracySec=1h
+RandomizedDelaySec=12h
+Persistent=true
+
+[Install]
+WantedBy=timers.target
--- a/system_files/shared/usr/libexec/bluefin-countme
+++ b/system_files/shared/usr/libexec/bluefin-countme
@@ -85,22 +85,37 @@ IMAGE_TAG=$(json_field image-tag)
 IMAGE_FLAVOR=$(json_field image-flavor)
 ARCH=$(uname -m)
 
-# image-name is the count. Without it there is nothing to report against, and a
-# guess would land in the wrong image's population.
-if ! sane "$IMAGE_NAME"; then
-    echo "no usable image-name in ${IMAGE_INFO_FILE}, not reporting" >&2
-    exit 0
+# What this system actually booted. Authoritative for the tag, and the fallback
+# for the name: image-info.json is a ublue convention and not every image in the
+# family ships one, but every bootc system knows its own ref.
+BOOTED_REF=$(bootc status --json 2>/dev/null |
+    jq -r '.status.booted.image.image.image // .spec.image.image // empty' 2>/dev/null || true)
+
+# ghcr.io/projectbluefin/bluefin-lts:stable -> bluefin-lts + stable
+#   digest-pinned refs carry no tag, so only the name is taken.
+REF_NAME=""
+REF_TAG=""
+if [[ -n "$BOOTED_REF" ]]; then
+    ref_body="${BOOTED_REF%@*}"
+    case "$ref_body" in
+        */*:*) REF_TAG="${ref_body##*:}"; ref_body="${ref_body%:*}" ;;
+        *:*)   REF_TAG="${ref_body##*:}"; ref_body="${ref_body%:*}" ;;
+    esac
+    REF_NAME="${ref_body##*/}"
 fi
 
 # The baked image-tag is the compose-time tag ("latest"); promotion retags the
-# same digest, so the tag a system actually follows only exists in bootc. Same
-# resolution order as /usr/bin/ublue-image-info.sh.
-BOOTED_REF=$(bootc status --json 2>/dev/null |
-    jq -r '.status.booted.image.image.image // .spec.image.image // empty' 2>/dev/null || true)
-case "$BOOTED_REF" in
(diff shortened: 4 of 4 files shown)


mrbobbytables (approved):

Danathar (changes requested):
The design is right and the write-up is careful — the client-side throttle, the opt-out-before-state, the "drop unusable metadata instead of sending it" rule are all the correct calls, and I'd like this to land soon because bluefin#1247, dakota#1561 and utah#145 are all queued behind it. But one thing doesn't work as shipped, and it's the part the PR calls out as the key fix.

**`bootc status --json` cannot run under `DynamicUser=yes`.** On bootc 1.16.7:

```
$ bootc status --json
```

`bootc status` requires root even for a read. The script swallows the failure (`2>/dev/null || true`), so `BOOTED_REF` is always empty, `REF_TAG` never sets `IMAGE_TAG`, and the tag falls back to the baked `image-tag` — which is `latest`. That is precisely the "every system reports `latest`" outcome the PR says it eliminates. The verification in the description was presumably run as root from a shell, where it works; the unit won't be.

Two ways out:

1. **Run the unit as root.** Drop `DynamicUser=yes`; keep `StateDirectory`, `ProtectSystem=strict`, `ProtectHome`, `PrivateTmp`, `NoNewPrivileges`, `RestrictAddressFamilies`, `SystemCallFilter` — they all still apply to a root service and the script only writes to its state dir. `bootc` then works as intended. This is the smaller change.
2. **Use `rpm-ostree status --json`** (`.deployments[] | select(.booted) | .["container-image-reference"]`), which works unprivileged and is what `/usr/bin/ublue-image-info.sh` — cited as the reference resolution order — actually uses. But it talks to rpm-ostreed over D-Bus, which needs `AF_UNIX`, and Dakota doesn't ship rpm-ostree, so you'd need the bootc path there anyway.

I'd go with (1).

**Worth verifying on a booted image either way:** `RestrictAddressFamilies=AF_INET AF_INET6` leaves out `AF_UNIX`. Bluefin's `nsswitch.conf` is `hosts: files myhostname mdns4_minimal [NOTFOUND=return] resolve [!UNAVAIL=return] dns`, and nss-resolve reaches systemd-resolved over a Unix (varlink) socket. It *should* fall through to `dns` against `127.0.0.53` when the socket is refused, but I couldn't run the transient-unit test from here to prove it. A quick `systemd-run -p DynamicUser=yes -p 'RestrictAddressFamilies=AF_INET AF_INET6' curl -sf https://countme.projectbluefin.io/` on a real system settles it; if it fails, add `AF_UNIX`.

Everything else I checked holds up: bucket boundaries match libdnf5 (0-6 / 7-27 / 28-139 / 140+), `ConditionPathExists=/run/ostree-booted` keeps it out of containers, the legacy `/etc/dakota-countme/disabled` path is honoured, and a failed ping leaves `lastrun` untouched.

Danathar (changes requested):
Re-reviewed at ee83bae against d46c4df. Thanks for the AF_UNIX/AF_NETLINK change, the separate stderr capture, `--data-urlencode`, and the 168-day bucket. That last one is right and my 140 was wrong: libdnf's `COUNTME_BUCKETS {2,5,25}` puts the bucket 3/4 boundary at 24 weeks.

**The main item is still open, and the PR now documents it the wrong way round.** `system_files/shared/usr/lib/systemd/system/bluefin-countme.service` still has `DynamicUser=yes`, and `docs/skills/image-registry.md` plus the comment above the bootc call in `system_files/shared/usr/libexec/bluefin-countme` now say `bootc status` needs no privileges and works under that sandbox. It does not. I ran the head script under a transient unit with the exact `[Service]` options from the file (bootc 1.16.7, on Bluefin):

```
sudo systemd-run --wait --pipe --collect -p DynamicUser=yes -p StateDirectory=bluefin-countme-probe \
  -p PrivateTmp=yes -p ProtectSystem=strict -p ProtectHome=yes -p NoNewPrivileges=yes \
  -p "RestrictAddressFamilies=AF_UNIX AF_NETLINK AF_INET AF_INET6" -p SystemCallFilter=@system-service \
  -E COUNTME_ENDPOINT=https://countme.example.invalid/metalink /usr/bin/bash < system_files/shared/usr/libexec/bluefin-countme

bootc status failed, booted image ref unavailable: error: Status: Preparing for write: Querying root privilege: This command must be executed as the root user
no usable image tag from bootc, reporting without one
```

So as shipped, every system reports without a tag. That is better than reporting `latest`, but the tag is the thing this PR exists to collect.

One correction to my earlier suggestion: dropping `DynamicUser=yes` on its own is not enough on bootc 1.16.7. As root with `ProtectSystem=strict` and `SystemCallFilter=@system-service`, `bootc status` fails with `Opening . with writable mount: Failed to remount: ExitStatus(unix_wait_status(159))`. 159 is SIGSYS: the `mount` syscall is not in `@system-service`. This combination works and keeps the rest of the sandbox:

```
# remove DynamicUser=yes
SystemCallFilter=@system-service @mount
```

I verified that with `systemd-run` and got the full JSON back. Please also fix the two doc and comment sentences that claim it works unprivileged.

**Second, still open from the automated review:** `bluefin-countme.timer` fires every 3 days and the script throttles at 7 days, so an always-on machine pings on day 9, not day 7 (the day 3 and day 6 firings are skipped). Either fire daily (`OnUnitInactiveSec=1d`) or lower the throttle to just under the timer period. Small change, but it is a built-in undercount.

Tests: `bats tests/test_bluefin_countme.bats` passes 16/16 at this head, and CI is green. Nothing else is outstanding from my side.

PR #1113, opened 2026-09-12
fix(cli): replace atuin and bash-preexec with starship and add tests

castrojo:
This fixes up the bluefin-cli ujust and strips down our payload more. 

### Summary
- Remove `atuin` and `bash-preexec` from `cli.Brewfile`
- Add `starship` to `cli.Brewfile` so `ujust bluefin-cli` installs and activates the Starship prompt
- Clean up leftover atuin comments and integrations in `bling.sh` and `bling.fish`
- Add Bats test coverage for `bluefin-cli` covering `fish`, `zsh`, unknown shells, and bundle/bling error aborts
- Add Bats test coverage for `ublue-bling` prompt decline behavior

cc @mrbobbytables


Danathar (approved):
Reviewed at 723ff3e. Checked out the head and ran the affected bats suites locally (tests/test_bling.bats, tests/test_native_recipes.bats, tests/test_bling_sh.bats, tests/test_bling_preexec_rearm.bats, tests/test_validate_brewfiles.bats): 34/34 pass in the first two, none fail in the rest. Grepped the whole tree for atuin and bash-preexec: the only remaining references are bling.sh's guarded sourcing of bash-preexec plus the bash-preexec-rearm.sh hook and its tests, all of which are no-ops when bash-preexec is absent (covered by the existing 'no bash-preexec installed' tests), and no docs or skills mention atuin. Two non-blocking notes for the record: starship is already in preinstall.d/system-cli.Brewfile and already initialised by bling.sh and bling.fish, so the prompt does not change for users; the real user-facing change is that ujust bluefin-cli no longer installs the atuin binary on fresh enables (existing installs keep it since brew bundle does not remove packages). The leftover bash-preexec sourcing in bling.sh could be cleaned up in a follow-up.

castrojo:
Status check: All CI checks are green and Danathar has approved. Waiting on a 2nd maintainer review (@hanthor, @ahmedadan, or @repires) to satisfy the +2 review requirement.

ahmedadan (approved):

hanthor (approved):

PR #628, opened 2026-06-11
test(scripts): add bats coverage for ublue-motd and profile.d scripts

castrojo:
## Summary

Adds `tests/test_ublue_motd.bats` covering `ublue-motd` and its profile.d companion scripts, which run in every interactive shell session.

### ublue-motd
- Template file existence check (graceful handling when template missing)
- envsubst expansion behavior
- tput terminal detection (non-tty path)
- glow rendering when glow is present vs absent
- Normal output path with fixture template

### ublue-motd.sh profile.d
- `$HOME/.config/no-show-user-motd` opt-out toggle — motd is suppressed when file exists
- Normal invocation path (opt-out file absent)

### umotd.sh profile.d
- Unconditional umotd invocation

## CI changes

- Adds shellcheck entries for motd scripts (if extensionless)
- Adds `Run bats (ublue-motd)` step

## Notes

unit-tests.yml is also modified by other test PRs from this sprint. Changes are additive; minor rebases may be needed.

Closes #590

PR #921, opened 2026-08-03
fix(systemd): remove rechunker local-fs ordering cycle

castrojo:
## Summary

- remove the `Wants=local-fs.target` and `After=local-fs.target` edges from `rechunker-group-fix.service`
- preserve ordering after `bootc-sysusers-shadow-sync.service` and before `systemd-sysusers.service`
- add a regression contract test and document the ordering invariant

The local-fs edges created an ordering cycle during upgrades from legacy rechunked images. The bridge unit must repair gshadow state before `systemd-sysusers.service`, without pulling in or waiting for `local-fs.target`.

Closes #918

## Validation

- `bash -n system_files/shared/usr/bin/rechunker-group-fix` — passed
- `git diff --check` — passed
- `bats tests/test_rechunker_group_fix.bats` — not run: `bats` is not installed in the environment
- `systemd-analyze verify ...` — not run: `systemd-analyze` is not installed in the environment
- `just check` — not run: `just` is not installed in the environment
- `pre-commit` — not run: `pre-commit` is not installed in the environment

PR #466, opened 2026-06-04
Update curated.yaml: added Ecosia browser in the browser section

k3nsh0:
## Description
This PR updates `curated.yaml` to include com.ecosia.Browser in the bazaar list. Ecosia is a social business based in Berlin that uses their profits to plant trees and fight climate change. They recently released [a Linux version](https://flathub.org/en/apps/org.ecosia.Browser) of their Chromium - based Ecosia browser. 

## Changes Made
- added com.ecosia.Browser to the curated list in `system_files/bluefin/etc/bazaar/curated.yaml`.

## Additional Context
I had a brief chat with @castrojo about this step during the latest Linux app summit. Tagging for review .

## Summary by CodeRabbit

* **New Features**
  * Added Ecosia Browser to the available browser options in the curated applications list.

PR #391, opened 2026-06-02
feat(hive): add bonedigger lifecycle bot and hive-status-sync

castrojo:
## Summary

Onboards `projectbluefin/common` into the hive agent workflow so issues can be queued, claimed, and worked by hive agents.

### What's added

- **`bonedigger.yml`** — reusable lifecycle bot from `projectbluefin/bonedigger` with `Common ⚙️` branding. Auto-creates all lifecycle labels (`queue/agent-ready`, `queue/claimed`, `status/approved`, etc.) on first run. Supports `/claim` `/unclaim` `/approve` slash commands.

- **`hive-status-sync.yml`** — hourly sync of hive agent formation status to the `todo.projectbluefin.io` project (`PVT_kwDOCCE0ds4BLZBC`). Reports common's build CI state and issue queue counts. Gracefully skips if `PROJECT_TOKEN` is not set.

### Manual steps needed after merge

- [ ] Add `PROJECT_TOKEN` secret to `projectbluefin/common` (same token used in dakota)
- [ ] Add `P0` label to `projectbluefin/common` (all other labels are auto-created by bonedigger on first issue open)

## Test plan
- Merge → bonedigger runs on first new issue
- After adding PROJECT_TOKEN: trigger `hive-status-sync` manually via workflow_dispatch to verify project status posts correctly

## Summary by CodeRabbit

* **Chores**
  * Added automated GitHub Actions workflows for enhanced project lifecycle management and status synchronization. These workflows enable continuous monitoring of issue activity, project health metrics, CI pipeline status, and contributor information, improving visibility into project operations.

PR #1116, opened 2026-09-13
fix(systemd): drop local-fs.target ordering from rechunker-group-fix.service

### LINKED ISSUE #918: fix(systemd): rechunker-group-fix creates local-fs ordering cycle on upgrades
## Problem

`rechunker-group-fix.service` now runs as an early-boot unit with `DefaultDependencies=no` and `Before=systemd-sysusers.service`, but it still declares:

```ini
Wants=local-fs.target
After=local-fs.target
```

On upgrades/rebases from legacy-rechunked Bluefin LTS images, this creates an ordering cycle through `local-fs.target` and `systemd-tmpfiles-setup-dev.service`. systemd deletes `local-fs-pre.target` to break the cycle, leaving `/var`/other fstab mounts unavailable and causing a boot hang.

Observed in [projectbluefin/bluefin-lts#391](https://github.com/projectbluefin/bluefin-lts/issues/391), with multiple affected users. Masking `rechunker-group-fix.service` was reported to dissolve the cycle, but simply removing it risks reintroducing the legacy-rechunk group-state boot failure.

## Evidence

Reported cycle:

```text
local-fs.target: Found ordering cycle on var.mount/start; has dependency on local-fs-pre.target/start, systemd-tmpfiles-setup-dev.service/start, systemd-sysusers.service/start, rechunker-group-fix.service/start, local-fs.target/start
local-fs.target: Job local-fs-pre.target/start deleted to break ordering cycle starting with local-fs.target
```

The current unit's early-boot ordering was introduced in common#530.

## Requested investigation

Determine the bootc/systemd-compatible ordering for this bridge unit. In particular, validate whether the `Wants=local-fs.target` / `After=local-fs.target` edges should be removed or replaced while preserving the pre-`systemd-sysusers` guarantee and coexistence with `bootc-sysusers-shadow-sync.service`.

Add a regression test or VM evidence proving that:

- legacy-rechunked upgrades still repair group/shadow state;
- normal systems do not acquire a local-fs ordering cycle; and
- bootc versions with `bootc-sysusers-shadow-sync.service` remain safe.

Danathar:
## The bug

`rechunker-group-fix.service` ships from here into `bluefin` and `bluefin-lts`, and it declares two things that cannot both be true:

- `Before=systemd-sysusers.service`
- `Wants=` / `After=local-fs.target`

Walking the stock edges between those two points shows why:

```
rechunker-group-fix.service   Before=  systemd-sysusers.service
systemd-sysusers.service      Before=  systemd-tmpfiles-setup-dev.service
systemd-tmpfiles-setup-dev    Before=  local-fs-pre.target
local-fs.target               After=   local-fs-pre.target
rechunker-group-fix.service   After=   local-fs.target             ← closes the loop
```

So every boot of a consuming image opens with an ordering cycle.

## Why the symptom is a coin toss

systemd breaks a cycle by deleting whichever job in the loop it reaches first, and that choice varies per boot and per image. Depending on the victim you get:

| Job deleted | What the user sees |
| --- | --- |
| `systemd-udevd` | 90 s device timeouts, `/var` never mounts, no home directories |
| `systemd-ask-password-console.path` | LUKS password agent never starts — encrypted volumes never come up after the initrd hands over |
| `systemd-sysusers` | missing system groups |
| `local-fs-pre.target` | assorted mount fallout |

One defect, several bug reports that look unrelated: projectbluefin/bluefin-lts#585, projectbluefin/bluefin-lts#466, projectbluefin/bluefin-lts#391 and #918.

> **Scope correction:** this PR was originally opened for projectbluefin/bluefin-lts#628 (LUKS hang on the current `:stable`). We now think that is **probably** a separate bug in the initrd — the Fedora 7.1.8 kernel's `CONFIG_MODPROBE_PATH` points at `/usr/bin/modprobe`, which CentOS Stream 10 userspace doesn't ship, so `dm-crypt` can't autoload — but that is not yet confirmed on the reporter's machine ([details and pending test](https://github.com/projectbluefin/bluefin-lts/issues/628#issuecomment-5655540119)). The cycle below is still real on that reporter's machine — their journal shows it — but it is not what stops their boot, and this PR should not be expected to fix #628.

**Confirmed on real hardware** since this PR was opened — the reporter in bluefin-lts#628 posted a journal from their own machine showing the cycle with yet another victim (the ESP's fsck job):

```
boot-efi.mount: Found ordering cycle on systemd-fsck@...832D-6772.service/start; has dependency on
  local-fs-pre.target/start, systemd-tmpfiles-setup-dev.service/start, systemd-sysusers.service/start,
  rechunker-group-fix.service/start, local-fs.target/start, boot-efi.mount/start
boot-efi.mount: Job systemd-fsck@...832D-6772.service/start deleted to break ordering cycle
```

## The change

**1. Order the unit the way its neighbours are ordered.** Match `systemd-sysusers.service` and bootc's own `bootc-sysusers-shadow-sync.service`:

```ini
After=systemd-remount-fs.service
After=bootc-sysusers-shadow-sync.service
Before=systemd-sysusers.service
```

No `local-fs.target` edges at all. The `systemd-remount-fs.service` anchor answers the "is `/etc` writable yet?" question raised in the #921 review — it is the same precondition sysusers itself relies on to write `/etc/passwd`.

**2. Drop the trailing `ExecStart=systemd-tmpfiles --create --remove --boot --exclude-prefix=/dev`.** With the unit now running before `local-fs-pre.target`, `/var` and `/tmp` are not mounted yet and that pass exits 65 on every single boot — this is the `rechunker-group-fix.service` failure line in every `ujust report`, and what @fabsah and @jlbauss were seeing. Nothing is lost by removing it: `systemd-tmpfiles-setup.service` runs the identical command later, after `local-fs.target` and after `systemd-sysusers.service`, i.e. after the groups repaired here exist.

**3. Document the boundary** in `docs/skills/submodule-boundary.md`, specifically why a downstream drop-in *cannot* fix this. systemd cannot reset `After=` or `Wants=` from a drop-in, and `After=` with an empty value is a no-op for dependencies (verified with `systemctl show -p After` on systemd 259). So the bluefin-lts#527 / bluefin-lts#603 style drop-ins only manage to *add* `Before=local-fs-pre.target`, producing a tighter cycle than the one they were meant to break.

> Supersedes #921 — same core edge removal, but without the unrelated riders it was blocked on, and adding the `systemd-remount-fs.service` anchor, the tmpfiles removal, and a transaction-level test. Happy to have this folded into #921 instead if that's preferred.

## Regression coverage

`tests/test_rechunker_group_fix.bats` adds three tests:

1. **Static ordering contract** — required edges present, no `local-fs*.target` edges, no `systemd-tmpfiles` ExecStart.
2. **Transaction-level check** — `systemd-analyze verify --root=<fixture> default.target` over a minimal fixture of the stock early-boot units (`systemd-remount-fs`, `systemd-sysusers`, `systemd-tmpfiles-setup-dev`, `systemd-udevd`, `local-fs-pre.target`, `local-fs.target`, `sysinit.target`). This lets systemd itself compute the boot transaction, and asserts no `ordering cycle` / `deleted to break` output.
3. **Guard against a dead fixture** — re-adds `After=local-fs.target` to a copy of the fixture and asserts systemd *does* report the cycle, so test 2 cannot silently pass on a broken fixture.

Run on Ubuntu 24.04 / bats 1.10 / systemd 255 (the CI runner environment):

```
# with this branch
ok 8  rechunker-group-fix.service: ordering contract matches systemd-sysusers.service
ok 9  rechunker-group-fix.service: systemd computes no ordering cycle for the boot transaction
ok 10 rechunker-group-fix.service: ordering fixture detects the local-fs.target cycle

# with the unit from main
not ok 8  rechunker-group-fix.service: ordering contract matches systemd-sysusers.service
not ok 9  rechunker-group-fix.service: systemd computes no ordering cycle for the boot transaction
# local-fs-pre.target: Found ordering cycle on systemd-tmpfiles-setup-dev.service/start
# local-fs-pre.target: Job systemd-tmpfiles-setup-dev.service/start deleted to break ordering cycle starting with local-fs-pre.target/start
```

## Validation performed

**Tests and linters**

- `bats tests/test_rechunker_group_fix.bats` — 10/10 pass (Ubuntu 24.04 container, systemd 255). Tests 8–9 fail against the unit on `main`.
- `shellcheck` on the script and the bats file; `just check`; `git diff --check`.
- `scripts/check-doc-links.sh`, `check-skill-index.sh`, `generate_skill_index.py --check` — all clean.

**Against the real consumer image**

`ghcr.io/projectbluefin/bluefin-lts:stable` (amd64 `sha256:33ba155b…`, kernel 7.1.8-200.fc44, systemd 257-33.el10), in a container with an Anaconda-style `/etc/fstab` + `/etc/crypttab` for a LUKS root. Ran `systemd-analyze verify --generators=yes` over `default.target`, `multi-user.target`, `graphical.target`, `cryptsetup.target` and `rechunker-group-fix.service`:

| Image | `ordering cycle` lines |
| --- | --- |
| as shipped | **16** |
| with this unit dropped in | **0** |

As shipped, one of those cycles runs through `cryptsetup.target` and deletes the LUKS password agent, which would leave any encrypted volume that still needs unlocking after switch-root without a prompt:

```
cryptsetup.target: Found ordering cycle on systemd-ask-password-console.path/start; has dependency on
  plymouth-start.service/start, systemd-udevd.service/start, systemd-sysusers.service/start,
  rechunker-group-fix.service/start, local-fs.target/start, zfs-mount.service/start,
  zfs-import.target/start, zfs-import-cache.service/start, cryptsetup.target/start
cryptsetup.target: Job systemd-ask-password-console.path/start deleted to break ordering cycle starting with cryptsetup.target/start
```

For comparison, the bluefin-lts#527/#603 drop-in measured in the same image: **14** cycle lines, versus 10 for the same anchor set without it — including `Job local-fs.target/start deleted`.

## Not performed

- **No real boot test.** Nothing here boots a VM or a LUKS machine. The evidence is the systemd transaction solver (`systemd-analyze verify`) run against the real image's unit set, plus the bats fixture. A lab boot of a consuming image — ideally an upgrade from a legacy rechunked image with a LUKS root, as in bluefin-lts#585 — is still needed before promotion, per the warning in the unit header.
- `pre-commit run --all-files` and `actionlint` were not run locally (not installed). The individual hook scripts listed above were run directly.

---

Closes #918
Refs projectbluefin/bluefin-lts#628 (scope corrected — not fixed by this PR)
Refs projectbluefin/bluefin-lts#585 projectbluefin/bluefin-lts#466 projectbluefin/bluefin-lts#391

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0114TUzBMEH13zPa2VVYwLdW


castrojo (approved):
Reviewed diff and verified ordering contract. Dropping Wants=/After=local-fs.target removes the boot ordering cycle on systemd-sysusers. Fixes projectbluefin/bluefin-lts#628 and common#918.

castrojo (dismissed):
LGTM, approving as second review to satisfy ruleset.

Danathar:
### Scope correction

This PR was opened as the fix for projectbluefin/bluefin-lts#628. It most likely isn't.

**What changed:** the reporter posted console output from the failing boot. It fails *inside the initrd*, before any of the units this PR touches ever run:

```
device-mapper: table: 252:0 crypt: unknown target type
```

**Current theory (not yet confirmed):** the Fedora `7.1.8-200.fc44` kernel is built with `CONFIG_MODPROBE_PATH="/usr/bin/modprobe"`, but CentOS Stream 10 only ships `/usr/sbin/modprobe`, so the kernel can't autoload `dm-crypt`. The reporter has been asked to test a kernel-arg workaround; write-up here: https://github.com/projectbluefin/bluefin-lts/issues/628#issuecomment-5655540119. If it holds up, that fix belongs in bluefin-lts, not this repo.

**What this means for this PR:**

- The ordering cycle it removes is still real — the same reporter's journal shows `rechunker-group-fix.service` in a cycle that deleted the ESP's fsck job.
- #918, bluefin-lts#585, #466 and #391 still stand.
- I've edited the description to drop `Fixes #628` and the "root cause of #628" wording, so this doesn't get merged on a false premise or auto-close that issue.

Danathar:
This is actually a little interesting...if true its exactly the kind of thing that could bite on an LTS version because it ships something different.

castrojo (approved):
LGTM: verified drop of Wants=/After=local-fs.target eliminates the ordering cycle on systemd-sysusers. CI checks green.

PR #1121, opened 2026-09-14
fix(chairlift): track rebranded projectbluefin/chairlift release

### LINKED ISSUE #1115: coordinate chairlift rebranding and schema check tracking to io.projectbluefin.chairlift
Tracking downstream integration of ChairLift rebranding from `frostyard/chairlift` to `projectbluefin/chairlift` (latest release `v0.12.2`).

Refs: projectbluefin/chairlift#18

eltorrero:
## Summary

Downstream coordination for the ChairLift rebrand (`frostyard/chairlift` -> `projectbluefin/chairlift`, `org.frostyard.ChairLift` -> `io.projectbluefin.chairlift`). Implements projectbluefin/chairlift#18 and closes projectbluefin/common#1115.

The rebranded release is live: `projectbluefin/chairlift` v0.12.2 produces GitHub releases directly (projectbluefin/chairlift#3), and `ublue-os/homebrew-tap` already carries a cask pinned to it (sha256 values verified against the release's `checksums.txt`).

### Changes

- **Cask source**: `chairlift.Brewfile` now taps `ublue-os/homebrew-tap` and casks `ublue-os/homebrew-tap/chairlift`, which pins the `projectbluefin/chairlift` v0.12.2 release. The old `frostyard/tap` cask is frozen at the pre-rebrand v0.10.1.
- **Schema gate**: `tests/check-chairlift-config` bumps `CHAIRLIFT_SCHEMA_REF` v0.10.1 -> v0.12.2 and repoints `UPSTREAM_RAW` to `https://raw.githubusercontent.com/projectbluefin/chairlift/...` in the same commit as the cask change, per the gate's own "bump together with the cask" invariant.
- **System files**: the desktop entry, three icons, and bootc polkit policy are renamed `org.frostyard.ChairLift.*` -> `io.projectbluefin.chairlift.*`. The policy and desktop entry are re-vendored from upstream v0.12.2 (the desktop keeps its only edit: the absolute wrapper `Exec=` path); the icons are byte-identical to v0.10.1, so they are pure renames. `bootc-update-stage` header comments follow the new action ID and policy path.
- **Handoff**: `brew-preinstall --external-chairlift` now also protects `ublue-os/homebrew-tap/chairlift` from OS-diet removal, alongside the historical unqualified, `frostyard/tap`, and `ublue-os/tap` spellings.
- **Tests**: `test_chairlift_config.py` updated for the new IDs, the new Brewfile, the v0.12.2 pin, and the v0.12.2 schema surface (8 new groups: `channel_group`, `update_all_group`, `sysupdate_updates_group`, `reset_group`, `dx_group`, `gaming_group`, `ai_group`, `troubleshooting_group`; new fields `ai_images`, `ai_model`). Bluefin's `config.yml` group choices are intentionally unchanged: every key it uses still exists in the v0.12.2 schema (verified live below).
- Pre-commit excludes, drift-workflow comment, and skill docs follow the rebrand.

### Verification

- `pytest tests/test_chairlift_config.py` — 29 passed
- `pytest tests/test_skill_docs.py` — 10 passed
- `python3 tests/check-chairlift-config` (live networked run) — "ChairLift config uses only keys upstream defines." against v0.12.2
- `bats tests/test_brew_preinstall.bats` — 48 passed, including the extended handoff test covering the new cask name
- cask sha256s in `ublue-os/homebrew-tap` match the v0.12.2 release `checksums.txt` exactly

### Coordination notes

- Dakota's `files/chairlift/install.sh` hard-checks the *old* Brewfile shape (`tap "frostyard/tap", trusted: true`) and overrides it at image build time. It needs a follow-up there: accept the new shape in its guard, and decide whether its override should also move from `ublue-os/tap/chairlift` to `ublue-os/homebrew-tap/chairlift`.
- Opting into the new fork-only groups (`sysupdate_updates_group`, `dx_group`, `gaming_group`, `ai_group`, ...) is a product decision requiring the matching helpers/backends on the image; this PR is packaging plumbing only.

— hive: backend=goose model=unsloth/Qwen3.8-27B
---
🐝 **Hive Agent**: `contributor` | **SHA:** `9823518`


Danathar (changes requested):
Verified the packaging side and it's clean: the `ublue-os/homebrew-tap` cask pins `projectbluefin/chairlift` v0.12.2 with the release URL; the vendored polkit policy is byte-identical to upstream `data/io.projectbluefin.chairlift.bootc.policy` at v0.12.2 (action id and `exec.path` match); the desktop entry differs only by the header comment and the absolute `Exec=`; both scalable icons have the same blob SHA as upstream. Ran `tests/check-chairlift-config` live (✓ against v0.12.2), `pytest tests/test_chairlift_config.py` (29 passed) and `bats tests/test_brew_preinstall.bats` (48 passed) at the PR head. Note no CI ran on this PR (empty checks rollup), so please push a fresh SHA to get a real run.

**Blocking: existing installs never migrate.** Bluefin doesn't pass `--external-chairlift`, so migration is whatever `brew-preinstall` + `brew bundle` do on the next boot, and for a machine that already has `frostyard/tap/chairlift` v0.10.1 installed that is:

1. `brew bundle` sees `cask "ublue-os/homebrew-tap/chairlift"` and asks `cask_installed?`. That check is by token — `cask_in_array?` falls back to `Utils.name_from_full_name(cask)`, i.e. `chairlift` — and `chairlift` *is* installed. It then asks `cask_upgradable?`, which compares the installed cask against its own tap's definition; `frostyard/tap` is frozen at 0.10.1, so it isn't outdated. Result: "Skipping install of ublue-os/homebrew-tap/chairlift cask. It is already installed."
2. The removal pass then sees `frostyard/tap/chairlift` dropped from the managed set, but `managed_name_present` also compares by token (`${managed##*/}`), matches it against the new cask, and skips the uninstall.
3. State gets stamped with the new hash, and every later boot early-exits.

So existing users keep the v0.10.1 binary indefinitely, while the image has removed `org.frostyard.ChairLift.bootc.policy` — the action that binary still requests. (pkexec falls back to the generic admin-auth action, so staging degrades to a password prompt rather than breaking outright, but the user-visible upgrade never happens.) Only fresh installs get v0.12.2.

Fix belongs in `brew-preinstall`, before bundling, something like: if `brew list --cask chairlift` reports the installed cask's tap as `frostyard/tap` (e.g. via `brew info --cask --json=v2 chairlift | jq -r '.casks[0].tap'`), `brew uninstall --cask frostyard/tap/chairlift` first so the bundle actually installs the new one — plus a bats case that stubs an installed frostyard cask and asserts the swap. dakota's `scripts/test_chairlift_migration.sh` is prior art for the shape.

Nit, not blocking: Homebrew's canonical tap name is `ublue-os/tap` (`brew tap-info ublue-os/homebrew-tap` normalizes to it), and that spelling was already in the handoff list. Using it in the Brewfile avoids a third spelling of the same cask in state files.

Danathar (approved):
Re-reviewed at 27d6e61. My blocker from the earlier review is fixed: `brew-preinstall` now uninstalls a ChairLift that was installed from `frostyard/tap` before `brew bundle` runs, so existing machines move to the `ublue-os/tap` v0.12.2 cask instead of staying on v0.10.1. The nit about the tap spelling is also done (`ublue-os/tap`).

What I checked:
- Read the full PR diff, including the new migration block, the bats cases and the doc changes.
- `ublue-os/homebrew-tap` `Casks/chairlift.rb` pins `version "0.12.2"` with the `projectbluefin/chairlift` release URL, matching `CHAIRLIFT_SCHEMA_REF`.
- `brew info --json=v2 --installed` on a real Homebrew install reports a `tap` for each installed cask, which is what the migration reads.
- Locally at this head: `bats tests/test_brew_preinstall.bats` 69/69, `pytest tests/test_chairlift_config.py` 29 passed, `tests/check-chairlift-config` passes against v0.12.2, shellcheck clean on `brew-preinstall`.
- Mutation: forcing `chairlift_migrated=1` makes 4 bats cases fail (60, 61, 63, 67), so the migration path is really under test.
- CI at this head: Build, Unit Tests, Validate PR, Validate Brewfiles and Validate ChairLift Config all succeeded. PR E2E composed the image, but the E2E suite job was skipped on this fork PR.

Danathar:
(Optional follow-up to my approval. Not blocking.)

Two of the new bats cases in `tests/test_brew_preinstall.bats` contain `! grep ...` checks that can never fail. In bash, `set -e` ignores a command that starts with `!`, so an `! grep` line only counts when it is the last line of the test.

- "leaves a merely-resolvable legacy chairlift installed elsewhere alone": neither the `! grep -q "brew uninstall --cask"` line nor the `! grep -q "brew untap frostyard/tap"` line is checked.
- "a Brewfile still on the legacy bare cask never migrates": the `! grep -q "brew uninstall --cask"` line is not checked.

To show it: I removed the `brew list --cask chairlift` gate from `brew-preinstall`, so the script now uninstalls even when ChairLift is not installed. The first test still passed. I then changed its two lines to `run ! grep -q ...` and the test failed as it should (`expected nonzero exit code!`). The production code is correct. The fix is simply to write `run ! grep` in those three places.

castrojo (approved):
LGTM. Verified the ChairLift rebrand packaging and migration logic:
- Upstream Cask pins projectbluefin/chairlift v0.12.2.
- Preinstall migration uninstalls frostyard legacy cask prior to bundle.
- Policykit, desktop entries, and icons match upstream specs.
- All required checks ('validate', 'Build and push image (x86_64)', 'Build and push image (aarch64)') are green.

PR #364, opened 2026-05-30
docs: note LTS changelog display issue (fixes #16)

hanthor:
Closes #16\n\nDocuments the LTS changelog display issue. Follow-up: add image-specific changelog function as suggested in the issue.

PR #1129, opened 2026-09-16
fix(bling): ship working ujust completions for bash/zsh/fish

### LINKED ISSUE #1126: ujust shell completion broken in bash & zsh, TAB completes files instead of recipes
### What happened, or what do you want?

Typing `ujust <TAB>` (or `ujust up<TAB>`) completes files in the current directory instead of `ujust` recipes, in both bash and zsh.

- `projectbluefin/bluefin#1171` — open, reproduces on `testing-20260915`.
- Bluefin LTS maintainer confirmed the same symptom on LTS (no LTS issue filed yet — this issue covers it).
- Precedent: `projectbluefin/dakota#358` ("No ujust completions", Bash and Fish) was fixed downstream in `dakota#364` by generating completions in Dakota's own `common.bst`. That workaround should become unnecessary once this is fixed at the source.

**Root cause** (verified in this repo):

1. `Containerfile` (build stage) generates the completions with a `sed`-rename:
   
   ```dockerfile
   RUN install -d /out/shared/usr/share/bash-completion/completions /out/shared/usr/share/zsh/site-functions /out/shared/usr/share/fish/vendor_completions.d/ && \
     just --completions bash | sed -E 's/([\(_" ])just/\1ujust/g' > /out/shared/usr/share/bash-completion/completions/ujust && \
     just --completions zsh  | sed -E 's/([\(_" ])just/\1ujust/g' > /out/shared/usr/share/zsh/site-functions/_ujust && \
     just --completions fish | sed -E 's/([\(_" ])just/\1ujust/g' > /out/shared/usr/share/fish/vendor_completions.d/ujust.fish
   ```
   
   Since `just` moved to dynamic `clap_complete`-based completion output, the emitted loader (e.g. a `source <(JUST_COMPLETE=bash ujust)` line plus a `complete ... _clap_complete_just just` binding) no longer survives the `sed s/just/ujust/` rename intact — the binding still registers against `just`, so `ujust` falls back to default file completion.

2. The generated files land in `/out/shared`, which the `ctx` stage copies **over** the checked-in tree (`COPY /system_files/shared ...` first, then `COPY --from=build /out/shared /system_files/shared`). Any tailored completion checked into `system_files/shared` is silently replaced by the broken generated shim.
3. `ujust` itself (`system_files/shared/usr/bin/ujust`) is a thin wrapper: `exec just --justfile /usr/share/ublue-os/just/00-entry.just "$@"`, and `00-entry.just` imports all recipe modules. So, recipe completion must resolve against the entry justfile, which the dynamic `just` loader doesn't do for the `ujust` command name.

---

## Proposed fix

- [ ] Remove or repair the `just --completions | sed` generator step in `Containerfile` so it no longer emits the broken `just`-bound loader.
- [ ] Ship tailored completions under `system_files/shared/` that complete `ujust` flags plus recipe names from `/usr/share/ublue-os/just/00-entry.just` (e.g. via `just --summary --justfile ...`), for **bash, zsh, and fish** (fish for parity with `dakota#364`).
- [ ] Add a build-time **content** assertion (not just file-existence, the broken shims also exist) so a regression fails the build, e.g. assert the installed bash completion binds `complete ... ujust` and the zsh file contains `#compdef ujust`.
- [ ] Add/extend BATS coverage that executes the completion scripts against a mock `just`/justfile and asserts recipe names are offered.

### Reproduction or evidence

Reproduced the mechanism on a live Bluefin system on latest testing images for Sept 15:

- Shipped files are 1-line dynamic-loader shims, **not** recipe completions:
  - `/usr/share/bash-completion/completions/ujust` (35 B): `source <(JUST_COMPLETE=bash ujust)`
  - `/usr/share/zsh/site-functions/_ujust` (129 B): `#compdef ujust` + `source <(JUST_COMPLETE=zsh ujust)` + call to `_clap_dynamic_completer_ujust`
  - `/usr/share/fish/vendor_completions.d/ujust.fish` (34 B): `JUST_COMPLETE=fish ujust | source`
- Bash, after sourcing the shipped file:
  - `complete -p ujust` → `no completion specification` (nothing registered for `ujust`; TAB falls back to file completion — the reported symptom).
  - `complete -p just` → registered there instead: the loader output ends with `complete ... -F _clap_complete_just just`, and its body re-invokes bare `"just" -- "${words[@]}"` with no `--justfile` pin.
  - Direct simulation (`COMP_WORDS=(ujust up)`, `COMP_CWORD=1`) yields an **empty** `COMPREPLY`.
- Zsh: `JUST_COMPLETE=zsh ujust` emits a script defining `_clap_dynamic_completer_just` (2 occurrences, 0 of the `_ujust` variant), but the shipped `_ujust` calls `_clap_dynamic_completer_ujust`, which results in undefined → silent no-op → file completion fallback.
- Root of the root cause: `just` 1.57's `--completions` no longer prints a full static script (bash now prints `eval "$(JUST_COMPLETE=bash just)"`), so the build-time `sed s/just/ujust/` never sees the `complete ... just` registration line. The rename strategy might not be effective anymore.

### Image and version

ghcr.io/projectbluefin/bluefin:testing-20260915

### Extra context

**Why here?**

- `bluefin#1236` (merged) shipped tailored bash/zsh completions plus existence checks in `20-tests.sh` and BATS tests, but Bluefin's `Containerfile` `ctx` stage lays `--from=common` **on top of** Bluefin's own `system_files`, so common's broken generated shims shadowed the fix and `#1171` still reproduces.
- `bluefin#1262` (open) is a second attempt at the same downstream fix and will hit the same shadowing.
- Fixing the generator (or replacing it with checked-in tailored completions) in `common` fixes bluefin, bluefin-lts, and utah at once, and lets Dakota drop its `common.bst` workaround.

---

## Acceptance criteria

- [ ] On a fresh boot of each consuming image (bluefin, bluefin-lts): `ujust <TAB>` lists recipes, `ujust up<TAB>` completes to `update` (or the matching recipe), in bash and zsh; fish completion present.
- [ ] `bluefin#1171` can be closed as fixed; LTS maintainer's report resolved with no LTS-side change.
- [ ] Dakota's `common.bst` completion workaround can be retired (follow-up in dakota repo, not this issue).

## References

- `projectbluefin/bluefin#1171` (open bug report)
- `projectbluefin/bluefin#1236` (merged downstream fix, shadowed, see repro comment on `#1171` from 2026-09-15)
- `projectbluefin/bluefin#1262`, `#1270` (open downstream retries)
- `projectbluefin/dakota#358` / `dakota#364` (prior art: fish + bash + zsh generation in downstream `common.bst`)
- This repo: `Containerfile` generator step; `ctx` overlay order (`COPY /system_files/shared` then `COPY --from=build /out/shared ...`); `system_files/shared/usr/bin/ujust` wrapper; `system_files/shared/usr/share/ublue-os/just/00-entry.just` imports.

KiKaraage:
## What does this change?

Change ujust completions from auto-generated `sed -E 's/([\(_" ])just/\1ujust/g'` to tailored files for both bash, zsh, and fish; plus add the ujust flag list & update `containerfile` skill.

## Why?

Closes #1126 

Related issues/PRs:
- projectbluefin/bluefin#1171 (bug report, open)
- projectbluefin/bluefin#1236 (downstream fix this ports for bash/zsh)
- projectbluefin/bluefin#1262, #1270 (redundant downstream retries)
- projectbluefin/dakota#358, dakota#364 (prior art, fish included)

## PR pipeline

```
opened ──▶ 4-review ──▶ approved ──▶ merged
```

> A maintainer reviews and approves; merge goes through the merge queue.
> Select `blocked` or `hold` to pause the work.

## Checklist

- [x] PR title follows Conventional Commits (`fix:`, `feat:`, `docs:`, `ci:`, `refactor:`, etc.)
- [x] `just check` passes (clean locally)
- [x] `pre-commit run --all-files` passes (passes locally)
- [x] Skill doc updated if the change affects agent-facing conventions or behavior (see `docs/skills/skill-improvement.md`): `docs/skills/containerfile/` updated, generation section rewritten
- [x] `AGENTS.md` / `docs/SKILL.md` / `docs/skills/` links remain valid
- [ ] CI is green after push: `gh run list --repo projectbluefin/common --limit 5`

Assisted-by: Muse Spark 1.3 via OpenCode
https://opncd.ai/share/HiYO5BRh


KiKaraage:
> [!NOTE]
> Muse Spark 1.3 typing on behalf of KiKaraage 

## Summary

`ujust <TAB>` completes files in the current directory instead of recipes, in bash and zsh, on every image that consumes this layer. Reported in projectbluefin/bluefin#1171 (open, reproduces on testing-20260915) and confirmed by the Bluefin LTS maintainer on LTS. Dakota hit the same class of bug before (dakota#358, worked around downstream in dakota#364).

Root cause: the `Containerfile` build stage generated completions with `just --completions <shell> | sed 's/just/ujust/'`. Since `just` moved to dynamic `clap_complete` loader output, that pipeline is structurally dead. `just --completions bash` now prints only `eval "$(JUST_COMPLETE=bash just)"`, so the build-time rename never sees the real registration line. The script that reaches the live system registers itself for the command name `just`, not `ujust`, and re-invokes bare `just` with no `--justfile` pin. Verified on a live testing-20260915 system: after sourcing the shipped file, `complete -p ujust` reports no completion specification, and a simulated `ujust up<TAB>` returns an empty match list. The zsh file fails the same way: it calls `_clap_dynamic_completer_ujust`, while the loader defines the `_just` variant.

This PR replaces generation with tailored completions that express what `ujust` actually is (`just --justfile <entry>`, not a renamed `just`) and adds a build-time gate so the broken output can never ship again.

## What changed

<details>
<summary>Per-file captions</summary>

- `system_files/shared/usr/share/bash-completion/completions/ujust`: tailored bash completion. Registers `complete -F _ujust ujust`, completes flags from the data file below, completes recipes via `just --summary --justfile <entry>`. Ported from bluefin#1236, plus test overrides and `-V`.
- `system_files/shared/usr/share/zsh/site-functions/_ujust`: tailored zsh completion. Recipes at position 2, static flags for dash-prefixed words, existing `_just` fallback kept for deeper positions. Ported from bluefin#1236, plus test overrides and flags.
- `system_files/shared/usr/share/fish/vendor_completions.d/ujust.fish`: new fish completion, recipes plus flags, `-f` so fish never falls back to files. Gives fish parity with dakota#364.
- `system_files/shared/usr/share/ublue-os/just/ujust-flags`: new single-copy flag list, one per line, read by all three completions at TAB time. When `just --help` gains flags, this is the only file that needs an edit.
- `Containerfile`: deleted the `just --completions | sed` generator. Added a `RUN` gate before the `ctx` stage that asserts each checked-in file binds `ujust`, contains no `JUST_COMPLETE` loader text, and that `/out/shared` ships nothing at those four paths (the overlay that used to shadow downstream fixes).
- `tests/test_ujust_completion.bats`: new, 26 tests, wired into the `Justfile` `test` recipe. Content assertions per shell, data-file shape tests, a no-embedded-lists test, and functional tests that source the real files against a mock `just` in bash, zsh, and real `fish --private`.
- `docs/skills/containerfile/`: updated the stale "ujust completion generation" section and red flag to describe hand-authored files, the gate, and the test overrides.

</details>

## Verification

- `bats tests/test_ujust_completion.bats`: 26 of 26 pass, including with real fish and zsh on the runner.
- Off-image run (entry justfile hidden in a mount namespace, simulating CI): 26 of 26 pass. The suite never depends on host image paths.
- Gate logic validated all three ways locally: correct files pass, a `JUST_COMPLETE` shim fails the build step, a file under `/out/shared` at the same paths fails the build step.
- Live behavior of the new files checked against real `just` 1.57.0 and the real entry justfile: `up` offers `update upgrade`, empty offers all recipes, `--l` offers `--list` and friends, `-V` offers `-V`, in all three shells.
- Not yet done: a full `podman build` and boot verification on a built image. That step still needs to happen before closing bluefin#1171.

KiKaraage:
## Instructions for consumer images

One shared completion serves every variant, because recipes resolve from each image's own `/usr/share/ublue-os/just/00-entry.just` at TAB time. No per-variant completion work is needed. Two rules apply to all consumers.

- Do not ship files at these four paths downstream. Bluefin's `Containerfile` copies its own `system_files` first (line 36) and then lays `--from=common` on top (lines 48-49), so common always wins there. Downstream copies never take effect and only confuse the next reader.
  - `usr/share/bash-completion/completions/ujust`
  - `usr/share/zsh/site-functions/_ujust`
  - `usr/share/fish/vendor_completions.d/ujust.fish`
  - `usr/share/ublue-os/just/ujust-flags`
- When `just` gains flags, update `ujust-flags` here in common. The change flows to every variant with no downstream edits.

Per image:

- Bluefin: close #1262 and #1270 as redundant, this supersedes both. The tailored bash/zsh files #1236 added under bluefin's `system_files/shared` are now dead weight under the overlay rule above, so a follow-up can delete them. Verify `ujust <TAB>`, `ujust up<TAB>`, and `ujust --l<TAB>` on the next testing build, then close #1171.
- Bluefin-lts: no action needed. Verify TAB completion on the next LTS build that picks up this common change.
- Dakota: `elements/bluefin/common.bst` installs common's `system_files/shared/usr` straight into the image (lines 93-94), but lines 106-113 then run the same broken `just --completions | sed` generator over the top, so Dakota is broken the same way today. After this lands, bump the common pin past it and delete lines 106-113. The tailored completions and the flags file then flow in with the `cp -r`, no replacement needed. Verify bash, zsh, and fish before and after.
- Utah: no action needed. Verify TAB completion on the next build.

## Follow-ups kept out of this PR

- Booting a built image and closing bluefin#1171 after verification.
- Deleting bluefin's now-shadowed copies from #1236, and in dakota bumping the common pin past this change and deleting the generator at `elements/bluefin/common.bst` lines 106-113.
- The repo style guide asks for `[[ ]]` while every bats file uses `[ ]`. The new test file follows the repo files. Settle the guide separately.

hanthor (approved):
This is the right fix in the right repo, and the build gate is the part that makes it stick.

## Fixing it here rather than downstream

The PR body lists five prior attempts — bluefin#1236, #1262, #1270, dakota#358, #364 — and that history is the argument for this change. `toggle-devmode` and friends live in common's `system.just`, so a fix applied in each consumer is a fix that has to be applied again for every consumer, forever. Ending the `sed -E 's/([\(_" ])just/\1ujust/g'` rewrite and shipping tailored completions is the version that can actually be finished.

The regex was never going to be right, either. It rewrites `just` → `ujust` on a `(`, `_`, `"` or space boundary across a generated completion script — so it depends on the exact shape of whatever `just --completions` emits, and silently produces something subtly wrong when upstream changes that shape. That's the class of bug where a green build ships broken completions.

## The Containerfile gate is the best part

```dockerfile
grep -qF 'complete -F _ujust ujust' /tmp/ujust-gate/ujust
grep -qF '#compdef ujust'           /tmp/ujust-gate/_ujust
grep -qF 'complete -c ujust'        /tmp/ujust-gate/ujust.fish
if grep -qF 'JUST_COMPLETE' ...; then echo "ujust completion is a just dynamic-loader shim" >&2; exit 1; fi
for f in ...; do if [ -e "/out/shared/${f}" ]; then echo "ujust completion shadowed by /out/shared/${f}" >&2; exit 1; fi; done
```

Two things it gets right that a lesser version wouldn't. The `JUST_COMPLETE` check catches a regression to a dynamic-loader shim — I appended `JUST_COMPLETE=1` to the bash file and confirmed the gate fires. And the `/out/shared` loop catches re-introduction of the *shadowing* that caused #1126 in the first place, rather than only checking the checked-in file is correct. That second one is the failure mode that would otherwise come back.

## Verification

`bats tests/test_ujust_completion.bats` → **26/26, 0 failures.**

Mutation-tested the bash completion rather than trusting the green run — all three caught:

| mutation | result |
|---|---|
| `complete -F _ujust ujust` → `… just` (break the binding) | **1 failed** |
| recipe lister returns nothing (`COMPREPLY=()` at :36) | **2 failed** |
| flags branch disabled (`COMPREPLY=()` at :28) | **2 failed** |

Reading `--version` etc. from a shipped `ujust-flags` data file rather than hardcoding a list in three scripts is the right call — one place to update, and the gate asserts `--version` is present in it.

## One caveat worth knowing

**7 of the 26 tests skip when zsh and fish aren't installed** — all the zsh and fish cases:

```
zsh completion lists recipes from the entry justfile   # skip zsh not installed
zsh completion offers flags for a dash-prefixed word   # skip zsh not installed
zsh completion falls back to recipes when flags missing # skip zsh not installed
fish completion offers matching recipes for a prefix   # skip fish not installed
... (4 fish cases total)
```

In my environment only bash has real coverage, and the run still reports 26/26 with nothing red. So the two shells this PR exists to fix beyond bash are only actually exercised if the CI runner ships zsh and fish — worth confirming, because if it doesn't, the zsh/fish completions are shipped untested while the suite looks fully green.

Two options, either fine: install zsh/fish in the test job, or assert the expected skip count so a silent drop to bash-only fails. The same pattern bit projectbluefin/finpilot#365, where three security-relevant cases skipped under root and the mutation survived unnoticed.

Approving — the gate means a regression here fails the build rather than shipping quietly, which is more than the previous arrangement could say.

---

hanthor (changes requested):
This is a good change and the diagnosis behind it is right. `just --completions bash | sed -E 's/([\(_" ])just/\1ujust/g'` was never going to survive contact with reality — it rewrote whatever `just` happened to emit, including internal identifiers, and it silently produced a different result every time upstream touched its completion templates. Replacing it with three tailored, readable files plus a shared `ujust-flags` data list is the right shape, and factoring the flags into a data file so the three shells stay in sync is a genuinely nice touch.

I want to single out the build-time gate, because it's the part I was most skeptical of and it mostly earned its keep. I extracted the `RUN` block verbatim and ran it against deliberately broken inputs rather than reading it and assuming. Eleven of twelve mutants were caught:

```
--- bash binds 'just' not 'ujust'              rc=1
--- zsh compdef just                           rc=1
--- fish complete -c just                      rc=1
--- --version dropped from flags               rc=1
--- bash regressed to JUST_COMPLETE shim       rc=1  ujust completion is a just dynamic-loader shim
--- zsh regressed to JUST_COMPLETE shim        rc=1  ujust completion is a just dynamic-loader shim
--- fish regressed to JUST_COMPLETE shim       rc=1  ujust completion is a just dynamic-loader shim
--- generator resurrected: /out/shared bash    rc=1  ujust completion shadowed by /out/shared/usr/share/bash-completion/completions/ujust
--- generator resurrected: /out/shared zsh     rc=1  ujust completion shadowed by /out/shared/usr/share/zsh/site-functions/_ujust
--- generator resurrected: /out/shared fish    rc=1  ujust completion shadowed by /out/shared/usr/share/fish/vendor_completions.d/ujust.fish
--- bash file emptied                          rc=1
```

The shadowing loop in particular is aimed at exactly the right thing — `COPY --from=build /out/shared /system_files/shared` at the end of the `ctx` stage really would overwrite the checked-in files, and the gate placement after every `/out/shared` write is correct. `Build` runs on `pull_request`, so this is a live PR gate, not decoration. Good work.

The blocking problem is that neither the gate nor the 26 new tests notice if the completions stop reading the flags file they ship.

`ujust-flags` is installed at `/usr/share/ublue-os/just/ujust-flags`, and each completion computes that path as a default beside the entry justfile. I broke that default in the bash completion only:

```diff
-    local flags_file="${UJUST_FLAGS_FILE:-${justfile%/*}/ujust-flags}"
+    local flags_file="${UJUST_FLAGS_FILE:-${justfile%/*}/just-flags}"
```

That ships a bash completion that will never find its flag list, so `ujust -<TAB>` offers nothing. Both defences stay green:

```
$ bats tests/test_ujust_completion.bats | grep -c '^not ok'
0
$ ./gate.sh   # the Containerfile RUN block, verbatim
GATE-PASS
```

The test that should have caught it is `completions read flags from the data file instead of embedding lists`, which does `grep -qF "ujust-flags" "${BASH_COMPLETION}"`. After the mutation, here is every line in that file matching it:

```
$ grep -nF 'ujust-flags' system_files/shared/usr/share/bash-completion/completions/ujust
3:# Flags come from the ujust-flags data file shipped next to the entry
```

The assertion is satisfied by the header comment. The prose survived; the code didn't. That's the same failure mode the old `sed` pipeline had — something that looks like it's checking the thing is actually checking a description of the thing — and it's worth fixing here precisely because this PR's whole purpose is to make the completions verifiably correct.

Two concrete fixes, both cheap:

- Anchor the greps to code rather than any occurrence. For bash, something like `grep -qE '^[[:space:]]*local flags_file=.*ujust-flags' "${BASH_COMPLETION}"`, and the equivalents for `_ujust` (line 12) and `ujust.fish` (line 21). The same applies to the entry-justfile assertion in `completions default to the image entry justfile`, which is a bare `grep -qF` today.
- Add a gate line tying the computed default to the installed path, so a rename on either side fails the build. The gate already has the install path in hand from the `COPY`; asserting that the string `ujust-flags` appears in an assignment (not a comment) in all three files would have killed this mutant.

While I was in there, the gate greps are also comment-insensitive in the other direction — commenting out the binding passes:

```
--- bash binding COMMENTED OUT                 rc=0  GATE-PASS
```

`grep -qE '^complete -F _ujust ujust'` instead of `grep -qF` closes that, and likewise for `^#compdef ujust` and `^complete -c ujust`.

Three non-blocking things worth knowing before you merge.

**The new test file never runs in CI.** You added it to the `Justfile` `test` recipe, but `.github/workflows/unit-tests.yml` enumerates bats files individually and the workflow is untouched by this PR:

```
$ git diff fbbb193..HEAD -- .github/workflows/
(empty)
$ grep -n "ujust" .github/workflows/unit-tests.yml
(no matches)
$ grep -o "tests/test_[a-z0-9_]*\.bats" .github/workflows/unit-tests.yml | sort -u | wc -l
21
$ ls tests/*.bats | wc -l
34
```

You're in company — thirteen files including `test_ujust.bats` are already unwired — but the effect is that only the Containerfile gate defends this in CI today, which makes the gate gap above matter more than it otherwise would. One step in `unit-tests.yml` fixes it.

**Seven of the 26 tests skip without zsh and fish**, so the zsh and fish behavioural cases reduce to static greps on any runner that lacks those shells. `unit-tests.yml` runs `ubuntu-latest` and installs only `bats`, so if you do wire the file in, add `zsh fish` to that `apt-get install` line or the shells you're actually fixing get the weakest coverage of the three.

**The static flag list has already drifted from `just`.** This is the cost of dropping generation, and it deserves to be a conscious trade rather than a surprise. Against `just 1.58.0`, seventeen real flags are missing from `ujust-flags`:

```
--alias-style --allow-missing --ceiling --chooser --cygpath --default-list
--dotenv-command --dotenv-filename --dotenv-path --group --indentation --json
--justfile-name --no-cache --tempdir --time --timestamp-format
```

I verified a sample are genuinely accepted by the installed `just` rather than parser artefacts, e.g. `just --json` emits the JSON dump and `just --group` errors with `a value is required for '--group <GROUP>'`. Nothing shipped is *invalid* — `--format` is a real alias for `--fmt` — the list is just stale. The gate only asserts `--version` is present, so drift is unbounded. A cheap guard would be a test that every line in `ujust-flags` is accepted by the installed `just`, which catches removals upstream even if it can't catch additions.

Also flagging that the `validate` check is currently red on this head (`mergeable_state: blocked`) — I didn't dig into it, but it'll need to go green for the merge queue.

The bash behavioural tests themselves are solid, for what it's worth. Renaming the entry justfile default, narrowing the dash glob so `-V` stops completing, dropping the flags-file existence guard, swapping `--summary` for `--list`, unbinding `complete`, and hardcoding the flag list all produced failures. It's specifically the flags-file path that slipped through.

---

Danathar (approved):
Ran the suite at `dbfd2e4` with real `zsh` and `fish` installed: 30/30 pass. Also pointed the bash completion at this machine's actual `/usr/share/ublue-os/just/00-entry.just`: `ujust tog<TAB>` offers `toggle-devmode toggle-tpm2 toggle-updates toggle-user-motd`, and `ujust --li<TAB>` offers the four `--list*` flags. Common's build (with the Containerfile gate) is green on the PR.

@hanthor's blocking point looks addressed in `dbfd2e4`: the test and gate greps are now anchored to the assignment lines (`^[[:space:]]*local flags_file=.*ujust-flags`, `^[[:space:]]*echo .*ujust-flags` for fish) and the binding checks use `^complete -F _ujust ujust$` / `^#compdef ujust$` / `^complete -c ujust `, so the "comment survives, code doesn't" and "binding commented out" mutants both fail now.

Two small follow-ups, neither blocking: `ujust-flags` is a snapshot — against `just 1.57.0` here it's missing 17 flags (`--alias-style --allow-missing --ceiling --chooser --cygpath --default-list --dotenv-command --dotenv-filename --dotenv-path --group --indentation --json --justfile-name --no-cache --tempdir --time --timestamp-format`) and lists `--format`, which isn't a just flag (`--dump-format` / `--evaluate-format` are). Worth a one-off refresh from `just --help` on the image's just version, but it doesn't change the fix.

Danathar:
@hanthor your changes-requested review from 09-17 is still the standing decision on this PR, but `dbfd2e4` (09-18) addressed it — the test and gate greps are now anchored to the assignment lines and the binding checks use `^complete -F _ujust ujust$` / `^#compdef ujust$` / `^complete -c ujust `, so the two mutants you built (comment-only match, commented-out binding) fail. I approved after re-running the suite with real zsh and fish; it needs your re-review to clear the block. Details in my review above.

castrojo (approved):
Approved after standards and specification review of head 055a887. The earlier completion-validation request is addressed by dbfd2e4; required checks are green.

castrojo (approved):

PR #1154, opened 2026-09-20
feat(countme): add dakota-scoped client and systemd units

castrojo:
Ship weekly anonymous countme timer, service, preset, and client script
in shared system files. Exits immediately for non-Dakota images and
reports Dakota systems to first-party countme deployment.

Assisted-by: Gemini via GitHub Copilot
Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


ahmedadan (approved):

Danathar (changes requested):
This adds `bluefin-countme.{service,timer}`, a preset, and `/usr/libexec/bluefin-countme` to `system_files/shared`, so the units ship on every image; the script itself exits 0 unless image-info.json's image-name starts with "dakota". Reviewed at 1d96ad8. bats passes 5/5 and CI is green. The buckets match libdnf (weeks 1 / 2-4 / 5-24 / 25+), stderr is not merged into the bootc capture, and the opt-out markers are checked before any state write. Five things are still off, most of them fixed on #1112 before it was closed.

**1. The bootc fallback cannot run under `DynamicUser=yes`.** I ran the head script under a transient unit with the PR's exact `[Service]` options on bootc 1.16.7, with `/run/ublue-os/booted-image` absent so it takes the fallback (URL swapped to an `.invalid` host so nothing real was pinged):

```
++ bootc status --json
+ booted_ref=
+ curl ... 'https://.../metalink?repo=dakota&tag=latest&flavor=main&gamemode=0&arch=x86_64&countme=1'
```

The same run as root gets `booted_ref=ghcr.io/ublue-os/bluefin-dx:stable` and `tag=stable`. On Dakota the primary source usually exists, so the fallback only matters when that file is missing; but then the script silently sends Dakota's baked image-tag, which is always `latest`. Either make the fallback work (drop `DynamicUser=yes`, keep `ProtectSystem=strict`, use `SystemCallFilter=@system-service @mount`, verified on #1112) or delete the fallback and report without a tag when the file is missing. `After=ublue-booted-image.service` on the service would also close the boot-time race.

**2. Timer cadence versus the throttle.** `OnCalendar=weekly` with `RandomizedDelaySec=12h` and no `FixedRandomDelay`, while the script skips any run less than 604800 s after `lastrun`. When this week's random delay is shorter than last week's, the run lands inside the 7-day window and is dropped without touching `lastrun`. Simulating that pair drops about 37% of firings (same shape as utah#145). `FixedRandomDelay=true` fixes it, or a throttle a bit under 7 days.

**3. Tag and flavor are interpolated straight into the URL.** #1112 moved to `curl -G --data-urlencode` after review; this PR went back to string building.

**4. Unit name and the Dakota handover.** The units are named the same as Dakota's existing `files/countme/bluefin-countme.{service,timer}`, and dakota's `common.bst` copies `system_files/shared/usr/` wholesale, so until dakota#1622 lands both elements install the same two paths (a BuildStream overlap; `project.conf` sets no fatal-warnings, so whichever stages last wins). After #1622 the files replace cleanly, but enablement changes: Dakota's `countme.bst` uses a static `timers.target.wants` symlink, this PR relies on `03-bluefin-countme.preset`, and I think Dakota only applies system presets on first boot (dakota's `swapfile.bst` says "presets only apply on first boot" for exactly this reason). If that is right, existing Dakota installs come up with the timer disabled after the switch. Ship the wants symlink in `system_files`, or have dakota#1622 keep one pointing at this timer. Also worth a line in the body: the epoch cookie moves from `/var/lib/dakota-countme-epoch` to the StateDirectory, so existing installs re-enter bucket 1 once.

**5. The Fedora ping is being dropped and the PR does not say so.** Dakota's current client pings mirrors.fedoraproject.org; this script only pings countme.projectbluefin.io, and dakota#1622 deletes the Fedora client. The body should state that Dakota stops appearing in Fedora's countme dataset, and `docs/skills/image-registry.md` (lines 105-117) still describes the Fedora path, `/usr/libexec/dakota-countme` and `/var/lib/dakota-countme-epoch`; #1112 had updated that file, this one does not.

Item 4's "presets only on first boot" is from Dakota's own comments and systemd's first-boot gate on `preset-all`, not tested on a Dakota VM. Happy to be corrected there.

castrojo:
Updated to resolve review feedback:
1. Dropped `DynamicUser=yes` in `bluefin-countme.service` while preserving sandboxing (`ProtectSystem=strict`, `PrivateTmp=yes`, `ProtectHome=yes`, `NoNewPrivileges=yes`, `RestrictAddressFamilies=AF_UNIX AF_NETLINK AF_INET AF_INET6`, `SystemCallFilter=@system-service @mount`), allowing the `bootc status --json` fallback to query root privilege when `/run/ublue-os/booted-image` is absent. Added `After=ublue-booted-image.service` to eliminate boot-time race.
2. Added `FixedRandomDelay=true` to `bluefin-countme.timer` and introduced a 24-hour slack margin in the client throttle to avoid dropped weekly firings.
3. Switched query construction to `curl -G --data-urlencode` in `bluefin-countme`.
4. Shipped `system_files/shared/usr/lib/systemd/system/timers.target.wants/bluefin-countme.timer` symlink so existing Dakota upgrades keep the timer active regardless of first-boot preset behavior.
5. Updated `docs/skills/image-registry.md` to document Dakota's first-party CountMe reporting, endpoint, and state path.
6. Isolated `bootc` stderr capture and added mocked bootc tests in `tests/test_bluefin_countme.bats`.

Tests run:
- `bats tests/test_bluefin_countme.bats` (6/6 passed)
- `just check` (syntax valid)
- `pytest tests/` (115/115 passed)

